Issue #1, opened 2021-12-05
Commutative Monoid Solver


TOTBWF:
When doing stuff with integers, it's pretty common to see some really gnarly goals involving nats.
For instance, this monster comes up when you try to define integer multiplication
```agda
a * x + b * y + (y + a * y + (x + b * x)) ≡ a * y + b * x + (x + a * x + (y + b * y))
```

A simple commutative monoid solver would make quick work of this, so it would be a nice-to-have.

## Design
The basic idea is to reflect expressions in some monoid into the following data type:
```
data Expr : Type ℓ where
  ε   : Expr
  _⊗_ : Expr → Expr → Expr
  [_↑] : M → Expr
```

Then, we can do a sort of NbE style proof where we interpret this into some semantic domain.
In our case, this will be sorted lists. Then, we shall define a sort of "readback" function that
takes a sorted list and reflects it back into `Expr`.

However, this poses some problems! In particular, we will need a notion of decidable ordering on `M`, which is no good.
However, we can circumvent this by using an ordering on `Term`! This requires quite a bit of machinery though, so it is probably best to split the work across several PRs.

TOTBWF:
We can also use the variable machinery to define solvers for commutative monoids/groups pretty easily.

TOTBWF:
The abelian group one is pretty much mandatory for abelian categories IMO

plt-amy:
Gonna close this as superceded by the nat and ring solvers

PR #5, opened 2021-12-06
Sum Types

TOTBWF:
# Description

This PR adds sum types, with a few little blurbs about why we should care about them. This is required for #1.

### DIFF
--- a/src/1Lab/Data/Sum.lagda.md
+++ b/src/1Lab/Data/Sum.lagda.md
@@ -1,0 +1,103 @@
+```agda
+open import 1Lab.Path
+open import 1Lab.Type
+
+open import 1Lab.Data.Dec
+
+module 1Lab.Data.Sum where
+```
+
+# Sum Types
+
+Sum types are one of the fundemental ingredients of type theory.
+They play a dual role to the [product type](agda://agda://1Lab.Types#_×_);
+if products allow us to state that we have elements of two types simultaneously,
+sum types allow us to state that we have an element of _one_ of two types.
+
+We use the notation `A ⊎ B` to hint at this type's set-theoretic analog: 
+the disjoin union.
+
+```agda
+infixr 1 _⊎_
+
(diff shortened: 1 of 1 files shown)


TOTBWF:
Not sure what's up with CI?

src/1Lab/Data/Sum.lagda.md
```diff
@@ -0,0 +1,103 @@
...
+
+open import 1Lab.Data.Dec
+
+module 1Lab.Data.Sum where
+```
+
+# Sum Types
+
+Sum types are one of the fundemental ingredients of type theory.
+They play a dual role to the [product type](agda://agda://1Lab.Types#_×_);
```
plt-amy, line 13: this is what's up with CI: the link says Type**s**, the module is just Type

plt-amy (changes requested):
Looks good other than the broken Agda link, thank you!

src/1Lab/Data/Sum.lagda.md
```diff
@@ -0,0 +1,103 @@
...
+```
+<!--
+```agda
+private variable
+    a b c d : Level
+    A : Type a
+    B : Type b
+    C : Type c
+    D : Type d
+-->
```
SquidDev, line 35: Does this --> need to be outside the code block? Sorry, not quite sure how Literate Agda works processes Markdown!

plt-amy: Oh, well spotted! It didn't cause a build issue because `-->` is also an Agda comment. @TOTBWF can you move it outside?

src/1Lab/Data/Sum.lagda.md
plt-amy: Also fundamental* on the line above

src/1Lab/Data/Sum.lagda.md
TOTBWF: Nice catch!

PR #6, opened 2021-12-06
Exponentials and order

uniwuni:
# Description
Adds exponentiation on the naturals, as well as some of its basic properties along with other order statements.

## Checklist

- [x] The imports are sorted (use `find -type f -name \*.agda -or -name \*.lagda.md | xargs support/sort-imports.hs`)

- [x] All code blocks have "agda" as their language. This is so that

- [x] Proofs are explained to a satisfactory degree; This is subjective,

However, it is **completely file** to only do one! For instance, I


plt-amy:
Looks good to me!

Issue #7, opened 2021-12-07
Split into base/properties modules


plt-amy:
The way `1Lab.Data` as a namespace is currently structured is very conducive to dependency cycles.

For instance, `1Lab.Data.Sum` depends on `1Lab.Data.Nat`, the former module needs a proof of commutativity of addition to prove sums are closed under h-levels. This is a problem if you want to use coproducts -- something very fundamental! -- in any module that might be imported from `Nat`.

Proposed solution: The following modules should all be split into `Base`/`Properties`, with the current module serving as a place to hang re-exports off of:

- [x] Nat
- [ ] Int
- [ ] List
- [ ] Sum

PR #9, opened 2021-12-08
Refactor: Split 1Lab.Data.Nat into base + props

plt-amy:
# Description

As part of addressing #7, splits the natural numbers into a `Base`
module and a `Properties` module. I'm making it a pull request
so others can weigh in on how the definitions got split.

## Checklist

- [x] The imports are sorted (use `find -type f -name \*.agda -or -name \*.lagda.md | xargs support/sort-imports.hs`)

- [x] All code blocks have "agda" as their language. This is so that

- [x] Proofs are explained to a satisfactory degree; This is subjective,


JacquesCarette:
I'm a huge proponent of small modules. They are necessary to prevent import cycles, but that's just a symptom.

The problem with large modules is their lack of conceptual clarity. The kind of mathematics in this library 'builds up' very naturally (pun intended), but only if you build up at the 'right' level of granularity.

The main point is that there is a real difference between a reusable library of mathematics and the 'chunking' one wants to do of the concepts for other purposes, such as pedagogy. Too bad literate Agda doesn't support transclusion.

plt-amy:
Hmm, you've given me an idea though:

Literate Agda doesn't support transcoding natively, but Pandoc _does_ support including other files using templates ("partials"). I wonder, if the Properties module got broken up into even smaller units (e.g. properties of `_+_` in a separate module from properties of `_*_`), whether partials could be used to present Nat in a “coherent” order

JacquesCarette:
Oh, that would be cool.  Having one order for compilation, and a different one for presentation, would be nice. I definitely would like that for some of my other work.

PR #11, opened 2021-12-11
wip: Cubical SIP

plt-amy:
# Description

This PR changes the definition of "standard notion of structure" to the definition
of "univalent structure" from [Internalizing Representation Independence with Univalence].

Unlike Angiuli et al.'s definition of univalent structure, and indeed Escardó's standard
notion of structure, the definition of `Structure` here talks about _homomorphisms_ more
generally, rather than limiting ourselves to homomorphic _equivalences_. This should
make it very simple to, in the future, prove that if `isSet (TypeWith S)`, then `TypeWith S`
assembles into a univalent category of S-structures.

As a nice benefit, this gives us combinators for piecing together structures, in addition
to maintaining the previous capability to add axioms, and netting us a result that
axioms can be transported through equivalences of underlying structures (`transferAxioms`)

[Internalizing Representation Independence with Univalence]: https://arxiv.org/pdf/2009.05547.pdf

In addition, this PR scraps the existing algebra pages (Monoid and Group) so that they can be
re-done in whatever style we agree to address #7 with.


JacquesCarette:
First: this is all really great. The code and explanations are very clear.

Second, meta: where is a good place to ask questions about some of the details? [No, this isn't a contradiction to the above - I just think that I am misunderstanding something, despite what seems to be a clear explanation!]

plt-amy:
The canonical place to ask questions (now!) is the Discussions page: https://github.com/plt-amy/cubical-1lab/discussions

The reason it only popped up today is that while I knew GitHub had a forum-like thing, but I couldn't for the life of me remember what it was called to set it up 😄 - @SquidDev to the rescue 🎉

plt-amy:
:tada:

PR #19, opened 2022-01-02
Feature: Deploy to GH Pages

plt-amy:
Adds a step to the CI workflow to deploy to GH pages. I'm just doing a PR to test whether it'll deploy from the PR, since that would be bad


plt-amy:
Let's hope it works on the main branch

PR #20, opened 2022-01-17
feat: Display type information on mouseover

SquidDev:
# Description
This annotates references to top-level variables with their type
information, which is then displayed when the variable is moused over.

 - Add a new Python script support/get-types.py. This is given a list of
   modules on stdin, and spins up an `agda --interaction-json` process,
   loading in _build/all-pages.agda, and querying types for each module.

 - Extends the ident -> href lookup code to also generate href -> ident
   lookups.

   We then use this map to find the type (via href -> (module, ident) ->
   type)), which is then added to the tag via the data-type attribute.

   Note, this requires us to generate the HTML in two steps - emitting
   md to _build/html1, where we then extract the ID map, before
   annotating and writing to _build/html.

 - Extend the existing hover JS code to also create a new element and
   position it when an identifier with data-type is hovered.

# Missing features
There's a couple of things probably worth doing before merging. I'm mostly putting this up on the off-chance folks have opinions about things.

 - [x] Strip "trivial" implicits from types. I couldn't see an option in Agda to do this for us, so it may be that the best solution is some regex and heuristics (e.g, remove all implicit level variables and their usages)

 - [ ] Generate types for Agda builtins. I'm currently getting the module list by listing the files in `src/`, which means Agda builtins (most notably `PathP`) don't have types. I don't know if it'd be better to list those in `_build/html0`?


plt-amy:
Seems like tooltips close to the bottom edge of the browser window will overflow the page:

screenshot

The black bar at the bottom is my Plasma status bar

plt-amy:
I implemented removal of implicits from the spines of types by making the Shakefile parse Agda code and remove the implicits straight from the AST, then print the code again :smile: 

[Here's a patch](https://files.amelia.how/Fwoa.patch)

SquidDev:
Oh, that's wonderful, thank you! I needed to add `--package Agda` in order for me to build this outside of Docker/Nix. Hoping that won't conflict with anything there?

I had a look at getting types for the Agda builtins, but it's a bit more work as they're not actually in scope when loading `_build/all-pages.agda` (as it doesn't import them directly).

I think the easiest thing to do here would be to emit `_build/html0`, and from that generate a `_build/everything.agda`, which we can load in Python instead. Don't know if that sounds reasonable?

plt-amy:
Adding `--package Agda` to the Stack invocation won't be a problem (but we should probably patch the README to mention that, in addition to Agda-the-program, you need Agda-the-Haskell-library when using Cabal).

As for types of builtins, maybe we can just add them to the `_build/all-pages.agda` generator? After all, they _are_ deployed pages, so logically they should be listed in "all-pages"..

plt-amy:
God damn it why do I have to approve every commit by hand >.<

plt-amy:
I've updated the container to be built with the Shakefile from this branch, let's hope it works!

PR #28, opened 2022-01-30
Replace Lua filter with Haskell code

SquidDev:
Instead of shelling out to Pandoc, we instead use the `Text.Pandoc` package. This allows us to remove the Lua filter, replacing it with normal Haskell code.

 - Use an oracle for compiling LaTeX snippets, meaning we get a cross-module cache (it's still terribly slow, alas).

 - Quiver diagrams are handled as before, just without writing the dependencies to a file - we use a WriterT instead, which allows us to run these jobs in parallel.

---

Nix definitions are almost definitely wrong, sorry about that - ended up needing far more packages than I was expecting.

Some other Shakefile stuff I think worth doing at some point - probably need to discuss this first :):
 - [ ] Detect code blocks without a language and complain (either a noisy log message or stop entirely)
 - [ ] Detect text which contains `<!--` or `-->` and barf.
 - [ ] Use `agda-reference-filter`/`agda-fold-equations` as libraries instead, just to avoid spinning up new processes.
 - [x] Replace `support/get-types.py` with some Agda code, so we can have an even more horrific Shakefile.


plt-amy:
CI dead because I removed the Python script. Of course, nix-collect-garbage has yeeted my statically-linked Agda, so it'll take a while for me to update the container

SquidDev:
I can't approve my own PR, but consider this approved.

PR #30, opened 2022-01-31
shake: Warn when resulting HTML contains <!-- or -->

SquidDev:
# Description

Prints a message like `[WARN] Data.Test.html contains misplaced <!-- or -->`. This isn't a fatal error right now, as I think it could get irritating for development. Happy to change this if preferred!

Ideally we'd include line numbers here, but that's going to be tricky to track and dubiously useful, as you're tracking line numbers in the HTML rather than the original Markdown/Agda. However, it should be pretty obvious when looking in the browser where the error is.

This is largely intended to track cases like the following:

~~~markdown
<!--

```agda
-- Some hidden code

-->
```

~~~


plt-amy:
I think I'd prefer if it were a fatal error, actually

PR #31, opened 2022-01-31
web: Include authors on each page

SquidDev:
Adds a "Written by: xyz" section to each page:

A screenshot of 1Lab.Path, with 'Written by Amélia Liao, Astra Kolomatskaia, Jonathan Coates and uni' on the bottom left.

It's a little dubious (git log isn't smart enough to exclude whitespace changes), but good enough. I think!

### DIFF
--- a/Shakefile.hs
+++ b/Shakefile.hs
@@ -33,6 +33,7 @@ import qualified Data.Text.Encoding as Text
 import qualified Data.Text.IO as Text
 import qualified Data.Map.Lazy as Map
 import qualified Data.Text as Text
+import qualified Data.Set as Set
 import Data.List.NonEmpty (NonEmpty((:|)))
 import Data.Traversable
 import Data.Digest.Pure.SHA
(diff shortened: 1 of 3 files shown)


plt-amy:
🎉

src/index.lagda.md
```diff
@@ -12,7 +12,7 @@ principle depends on [_universes_](agda://1Lab.Type), [_propositional
 equality_](agda://1Lab.Path) (paths) and [_equivalences_](agda://1Lab.Equiv). 
 In addition to the hyperlinked "web of concepts" provided by the Agda
 code, there is a short introduction to homotopy type theory: **[Start
-here](1Lab.intro.html)**.
+here](1Lab.Intro.html)**.
```
plt-amy, line 15: What happened here?

plt-amy (changes requested):

src/index.lagda.md
SquidDev: I also noticed the link to 1lab instead of 1Lab, but fixed the wrong thing instead. I cannot explain myself.

plt-amy: haha

PR #42, opened 2022-02-20
Feature/Coyoneda

TOTBWF:
# Description

This PR proves the Coyoneda Lemma, and also defines some machinery needed to state it.


plt-amy:
code/text look good to me 👍

PR #43, opened 2022-02-21
refactor: use kebab-case everywhere

plt-amy:
pained screaming


ralsei:
Indentation's broken in a few places after this. Namely, for example, in `Relation.Order`:
```agda
deform x i .preorder .propositional =
  isProp-isProp (order .preorder .propositional)
                (x .preorder .propositional)
                i
```
gets refactored into
```agda
deform x i .preorder .propositional =
  is-prop-is-prop (order .preorder .propositional)
                (x .preorder .propositional)
                i
```
when the bottom should be re-indented to fix the naming.

That, and also I don't like the name `is-prop-is-prop`. But I didn't like `isProp-isProp` either, and I don't have any other ideas for what that should be named.

SquidDev:
Places which still need updating. Will continue to edit this as I go through files.

 - [ ] `iso-prop`? https://github.com/plt-amy/cubical-1lab/blob/a5f2e474a3d207df74c8038394417cc68e6b364b/src/1Lab/Counterexamples/IsIso.lagda.md#L41
 - [ ] Indentation needs updating? https://github.com/plt-amy/cubical-1lab/blob/a5f2e474a3d207df74c8038394417cc68e6b364b/src/1Lab/Equiv/HalfAdjoint.lagda.md#L148-L149 and https://github.com/plt-amy/cubical-1lab/blob/a5f2e474a3d207df74c8038394417cc68e6b364b/src/1Lab/Equiv/HalfAdjoint.lagda.md#L189-L190

plt-amy:
Fixed, thanks so much!

plt-amy:
`is-prop→PathP` probably needs to become `is-prop→path-p`. same for is-set→SquareP

src/1Lab/HIT/Truncation.lagda.md
```diff
@@ -62,11 +62,11 @@ whenever it is a family of propositions, by providing a case for
...
   go (inc x) (inc x₁) = work x x₁
   go (inc x) (squash y y₁ i) =
-    is-prop→PathP (λ i → pprop (inc x) (squash y y₁ i))
+    is-prop→path-p (λ i → pprop (inc x) (squash y y₁ i))
                   (go (inc x) y) (go (inc x) y₁) i
 
   go (squash x x₁ i) z =
-    is-prop→PathP (λ i → pprop (squash x x₁ i) z)
+    is-prop→path-p (λ i → pprop (squash x x₁ i) z)
                   (go x z) (go x₁ z) i
```
SquidDev, line 70: B-b-b-but I'd checked these files already!

plt-amy: that's the only one I messed up, I looked at all the others!

PR #53, opened 2022-04-21
Couple web improvements

ncfavier:
Currently [1Lab.Counterexamples.IsIso](https://1lab.dev/1Lab.Counterexamples.IsIso.html) doesn't fit horizontally on my screen, and I can't even scroll it:

This fixes that, and also makes the 🧊 link back to the index so that you don't have to scroll to the bottom of the navigation bar.

### DIFF
--- a/support/web/agda-cats.scss
+++ b/support/web/agda-cats.scss
@@ -59,7 +59,6 @@ article {
   padding-top: 2em;
   padding-bottom: 2em;
   margin-top: 0 !important;
-  width: 85ch;
   max-width: 100%;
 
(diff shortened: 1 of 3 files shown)


plt-amy:
Could you add an `id` to the `<a>` containing the cube logo, and add a rule

```
a[href]#cube.hover-highlight {
  background-color: unset;
}
```

to the stylesheet? otherwise the cube lights up 🥴

ncfavier:
Done

plt-amy:
Thank you!

PR #54, opened 2022-04-24
defn: reasoning combinators for functors

TOTBWF:
# Description

This PR adds a handful of reasoning combinators for functors ala `Cat.Reasoning`. I've also stolen `pushl`/`pushr` from `agda-categories`, as those can be useful.

# Notes
Coming up with these names is hard! I'm definitely open to suggestions.


plt-amy:
Can we steal Jacques' `⟩∘⟨refl ⟩` stuff too?

PR #57, opened 2022-04-28
Fix link to overture on topos theory; fix typo

tomdjong:
1. The link to the overture on topos theory now points to `Topoi.Base.html` (and removes an excess `.html` in the link). 
2. The typo fix changes `\Sets` to `\sets` so that it renders correctly.


plt-amy:
Thank you!

PR #60, opened 2022-05-02
Allow building on Stack again

SquidDev:
Untested on Nix I'm afraid.


plt-amy:
Still works on Nix. Missing while we're at it:

- Does building `agda-typed-html` work with Shake? If so, can you add that to the README?
- We should probably pretty-print the Agda errors... Ping me in the morning and I'll figure out how it works

PR #63, opened 2022-05-05
web: add navbar inline footnote checkbox (not tested)

person-with-a-username:
# Description

Add a footnote checkbox to the navbar so that people with small screens can toggle inline footnotes.

I generalized `unfold.js` to allow for multiple inline footnote checkboxes. However, I did this all from the GitHub web editor and didn't test it, so it may not work. Also, the code is very bad, so maybe it can be improved or made to be more consistent with other code?

## Checklist

*(Hopefully? I didn't edit any Agda code.)*

- [ ] The imports are sorted (use `find -type f -name \*.agda -or -name \*.lagda.md | xargs support/sort-imports.hs`)

- [ ] All code blocks have "agda" as their language. This is so that

- [ ] Proofs are explained to a satisfactory degree; This is subjective,

However, it is **completely fine** to only do one! For instance, I

### DIFF
--- a/support/web/unfold.js
+++ b/support/web/unfold.js
@@ -66,14 +66,20 @@ document.addEventListener("DOMContentLoaded", () => {
   });
 
   if (footnotes) {
-    const fnctl = document.getElementById("footnote-control");
-    fnctl.style.display = "flex";
-    const selected = document.querySelector("span#footnote-control > input");
+    const fnctls = document.querySelectorAll("span.footnote-control");
+    fnctl.forEach(fnctl => fnctl.style.display = "flex");
+    const selecteds = document.querySelectorAll("input.footnote-control");
 
-    selected.checked = unfold_footnotes;
-    selected.onchange = () => {
-      unfold_footnotes = selected.checked;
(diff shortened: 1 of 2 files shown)


support/web/unfold.js
```diff
@@ -66,14 +66,20 @@ document.addEventListener("DOMContentLoaded", () => {
   });
 
   if (footnotes) {
-    const fnctl = document.getElementById("footnote-control");
-    fnctl.style.display = "flex";
-    const selected = document.querySelector("span#footnote-control > input");
+    const fnctls = document.querySelectorAll("span.inline-footnotes");
+    fnctl.forEach(fnctl => fnctl.style.display = "flex");
```
plt-amy, line 70:
```diff
-    fnctl.forEach(fnctl => fnctl.style.display = "flex");
+    fnctls.forEach(fnctl => fnctl.style.display = "flex");
```

plt-amy (changes requested):

plt-amy:
Thanks! The CI failure is unrelated.

PR #66, opened 2022-05-06
fix nix-build command

Ryunaq:
Just a silly typo that confused me a bit ;p


ncfavier:
Needs to be fixed in the paragraph just above as well.

ncfavier:
This was done in https://github.com/plt-amy/1lab/pull/124

PR #67, opened 2022-05-07
🤖 🔢 Implement a Nat Solver

TOTBWF:
# Description

This PR implements a nat solver, as well as some reflection machinery.

## Notes
I got burnt out defining alpha equality for terms. Perhaps someone else wants to do it, but I cannot be bothered! We have everything we need for our purposes here anyways.


plt-amy:
🙏🙌

plt-amy (approved):
👍

PR #69, opened 2022-05-09
Basic citation support

### LINKED ISSUE #68: Citations
As the 1Lab grows, we should be mindful of properly citing our sources! For instance, #67 pulls quite a bit from multiple sources, and it would be great if we had a nice way of giving credit. Perhaps @SquidDev has some thoughts?

SquidDev:
Closes #68. See the notes in the commit message for actual implementation details.

Rendered view:

"A screenshot of the 1Lab.intro page, with a "References" block

CC @plt-amy and @TOTBWF - don't know if you have opinions on citation styles?

It might also be nice to move our `:target` highlight class to apply to all links, so people know which reference they're looking at.


src/1Lab/intro.lagda.md
```diff
@@ -347,12 +347,10 @@ _$\beta$-reducible expression_, but since we're not really interested in
 the theory of programming languages here, this terminology will not come
 up.
 
-[^newoldlang]: A [survey paper] of POPL proceedings by Guy L. Steele
+[^newoldlang]: A survey paper of POPL proceedings by Guy L. Steele [@Steele2017]
```
plt-amy, line 350: I think this line is over 72 characters but I can't quite tell — can you wrap it?

support/shake/app/Main.hs
```diff
@@ -33,10 +34,12 @@ import qualified System.Directory as Dir
 import Text.HTML.TagSoup
 import Text.DocTemplates
 
+import Text.Pandoc.Citeproc
 import Text.Pandoc.Filter
 import Text.Pandoc.Walk
 import Text.Pandoc
 import Text.Printf
+import qualified Citeproc as C
```
plt-amy, line 42: can you change this to `as Cite`?

plt-amy (changes requested):
Tiny tiny quibble

plt-amy (commented):
Nice, but also, yikes.

TOTBWF:
Awesome, thank you so much ❤️

Issue #72, opened 2022-05-11
Rewrite build-diagram so it doesn't depend on rubber


plt-amy:
Getting rid of the Python dependency is good for the soul and good for an 80MiB reduction in Docker image size. @SquidDev thoughts?

plt-amy:
Ideally we'd combine this with #71 so the Shakefile calls `pdflatex` and `pdftocairo` by itself.

Issue #71, opened 2022-05-11
Refactor Shakefile into a bunch of Haskell modules


plt-amy:
Our Shakefile is 600 lines and it's an affront to God. @SquidDev thoughts?

plt-amy:
Some concrete wants:

- Bring agda-reference-filter and agda-fold-equations into the 1lab-shake process
- The `buildMarkdown` action needs to be chopped up into a million tiny pieces
- The comment headers provide pretty good splitting points for the modules already
- Add a comment explaining the difference between `patchInlines` and `patchInline`
- Haddock _everything_

SquidDev:
> Bring agda-reference-filter and agda-fold-equations into the 1lab-shake process

By this, do you mean link as libraries, or just copy the source wholesale into the tree?

plt-amy:
The latter, they're both pretty short

plt-amy:
Also we need a jazzier name for our build system,, like GHC has "Hadrian"

PR #80, opened 2022-05-17
Remove need for _build/html1 directory

SquidDev:
- Remove Rubber dependency (#72)
 - Drive `agda://` links using the Agda type dump (technically part of #71).

Confirmed this produces near-identical output to main. The only difference is in `static/links.json` - as we build this file later (and handle `agda://` links earlier), those are now included in the link map.


SquidDev:
I was sure I'd used that pun already, but git history says otherwise.

PR #94, opened 2022-05-20
Search for headings

### LINKED ISSUE #73: Search for headings
We should augment our call to Pandoc so it also yeets a list of all headings. Then we should add these headings to the search prompt.

SquidDev:
A screenshot of a search for "spaces". The results contain 5 headers, each with some summary text below.

Closes #73. As mentioned on IRC, it's not perfect - really you want a different search metric for headings/text and code/Agda identifiers, but then merging the results of the two is very hard. But I think a good starting place.

### DIFF
--- a/support/shake/1lab-shake.cabal
+++ b/support/shake/1lab-shake.cabal
@@ -49,6 +49,7 @@ executable shake
       , Shake.LinkGraph
       , Shake.LinkReferences
       , Shake.Markdown
+      , Shake.SearchData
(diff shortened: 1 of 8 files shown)


plt-amy:
Figured out what the problem was: It wasn't creating the _build/search directory

plt-amy:
Well ain't that annoying

plt-amy:
I'll just take a leap of faith and rebuild since it works on my machine...

PR #102, opened 2022-07-07
`^op` → `\op` in LaTeX

ncfavier:


SquidDev:
I don't think `\op` is a valid LaTeX term. At least KaTeX and MathJax don't appear to support it (you get $\op$).

plt-amy:
It's one of our macros

plt-amy:
Thanks for the contribution! I keep messing this up haha

SquidDev:
Ahh, thanks! Sorry, I'm a plonker.

PR #104, opened 2022-07-23
wip: displayed category solver

plt-amy:
Solver for associativity/unit for displayed categories. Still need to wire it up to reflection machinery


plt-amy:
@TOTBWF CI failure is because of [this file](https://github.com/plt-amy/1lab/blob/9de8d25e5c42f043f759cf6ae69241ccfdc03248/src/Cat/Instances/Delooping.lagda.md#L46) still having a reference to CategoryNames

plt-amy:
Problem now is that `Algebra.Group.Ab` uses the monoid solver in two places

TOTBWF:
Yeah, I have most of a new shiny group solver written, will add tonight.

PR #108, opened 2022-07-26
Implement a Functor Solver

TOTBWF:
# Description

This PR implements a little functor solver. There's no commentary here, as it's not very enlightening. We should have a chat about commentary for solvers though! A style guide for solvers would also be a good addition :)


plt-amy (approved):
Can you `runghc support/sort-imports.hs src/Cat/Functor/Solver.agda`?

PR #111, opened 2022-07-27
Implement a Monad Solver

TOTBWF:
# Description

This PR implements a monad solver, and also fixes a reflection oversight in the original functor solver reflection code.


TOTBWF:
Ah, I actually missed a case where `M₁ (mult.η X) ∘ M₁ (unit.η (M₀ X)) ≡ id` doesn't get solved. I think this just means we need to push kmap frames a bit deeper.

plt-amy (approved):

PR #116, opened 2022-07-29
fix typo: refer -> reserve

JakobBruenker:
In a paragraph with lots of "refer"s, one "reserve" was caught in the crossfire


plt-amy:
Thanks!

PR #122, opened 2022-08-13
Update Discord links

szumixie:


plt-amy:
Thanks! I ended up doing it myself to use a permanent invite link, though

PR #123, opened 2022-08-16
add explanation of `ap-comp-path` as a uniqueness property

ncfavier:
# Description

An alternative explanation for the proof of `ap f (p ∙ q) ≡ ap f p ∙ ap f q`.

Also, could you please add `Naïm Favier <n@monade.li>` to the mailmap? (I've been using my full name inconsistently which makes me appear twice on some pages)

## Checklist

- [X] The imports are sorted (use `find -type f -name \*.agda -or -name \*.lagda.md | xargs support/sort-imports.hs`)

- [X] All code blocks have "agda" as their language. This is so that

- [X] Proofs are explained to a satisfactory degree; This is subjective,

However, it is **completely fine** to only do one! For instance, I

### DIFF
--- a/src/1Lab/Path.lagda.md
+++ b/src/1Lab/Path.lagda.md
@@ -585,65 +585,6 @@ inspect x = x , refl
 ```
 -->
 
-## Functorial Action
-
-In HoTT, every function behaves like a funct**or**, in that it has an
-action on objects (the actual computational content of the function) and
-an action on _morphisms_ --- how that function acts on paths. Reading
-paths as identity, this is a proof that functions take identical inputs
-to identical outputs.
-
-```agda
-ap : ∀ {a b} {A : Type a} {B : A → Type b} (f : (x : A) → B x) {x y : A}
-   → (p : x ≡ y) → PathP (λ i → B (p i)) (f x) (f y)
-ap f p i = f (p i)
-```
-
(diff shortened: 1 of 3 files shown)


plt-amy:
Hmmm.. can you try replacing the proof of ap-comp-path with your version to see if CI passes? Makes for a better proof imo. I'll fix the mailmap when I manage to shuffle out of bed

plt-amy:
feel free to axe it

ncfavier:
Alright here's a suggestion: we move `ap-comp-path` back into `1Lab.Path`, but since it needs uniqueness we move the entire "functorial action" subsection into its own section *after* the section about composition (gotta have a category before we can talk about functors!). And then the example in "syntax sugar" needs `ap` so we can move that section after functorial action. Does that sound good?

plt-amy:
Sounds great, go for it!

plt-amy:
[gonna use this PR to test the webhook, sorry for the noise]

ncfavier:
Some identifiers seem to not get linkified in `1Lab.Path.Groupoid`:



also `hfill` and `Path`. Any idea why? `ap-refl` and `ap-sym` were links before this PR.

plt-amy:
No references to those identifiers in the module, I assume?

ncfavier:
Oooh, so that's what the `_ = ap-refl` lines were for 🤦‍♂️

PR #124, opened 2022-08-16
chore: fix the Nix build

ncfavier:
Makes `1lab-shake` more self-sufficient by wrapping it with its Node dependencies so we don't have to link node_modules into the current directory.

Also sets `--git-dir` explicitly on git calls to avoid the `safe.directory` hack.

Subsumes https://github.com/plt-amy/1lab/pull/66.

I would have liked to use [flakes](https://nixos.wiki/wiki/Flakes) to make dependency management easier, but this is pretty much dead in the water because of https://github.com/NixOS/nix/issues/6900. That's no big deal though.

I'd also like to convert the CI to use Nix instead of a Docker container. We'd still be caching `_build` of course, and only using Nix to build the shakefile. The main advantage of this (apart from the obvious "yeeting Docker") is that we could easily set up a [binary cache](https://www.cachix.org/) so that users don't need to build Agda locally after it's been built by CI.

However this isn't quite ready yet: my tests pull a lot more paths from caches than the size of the docker image, so there are probably some runtime dependencies that aren't properly getting killed. I'll look into it later.

(CI failing expectedly because it's still using the old docker image)

### DIFF
--- a/support/shake/app/Main.hs
+++ b/support/shake/app/Main.hs
@@ -140,7 +140,7 @@ rules = do
     getDirectoryFiles "support/web/js" ["**/*.ts", "**/*.tsx"] >>= \files -> need ["support/web/js" </> f | f <- files]
 
     let inp = "support/web/js" </> takeFileName out -<.> "ts"
-    command_ [] "node_modules/.bin/esbuild"
+    command_ [] "esbuild"
       [ "--bundle", inp
       , "--outfile=" ++ out
       , "--target=es2017"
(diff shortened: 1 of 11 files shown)


ncfavier:
Here's a complete successful workflow run on my fork using the new docker image: https://github.com/ncfavier/1lab/runs/7868434335?check_suite_focus=true

plt-amy:
Thanks!

support/shake/app/Main.hs
```diff
@@ -140,7 +140,7 @@ rules = do
     getDirectoryFiles "support/web/js" ["**/*.ts", "**/*.tsx"] >>= \files -> need ["support/web/js" </> f | f <- files]
 
     let inp = "support/web/js" </> takeFileName out -<.> "ts"
-    command_ [] "node_modules/.bin/esbuild"
+    command_ [] "esbuild"
```
SquidDev, line 143: Doing this will breaks building outside of Nix (so with Cabal/Shake), as `esbuild` isn't on the path.

I'm not sure what the right fix here is - maybe worth adding a `nodeCommand = command_ [AddPath "node_modules/.bin"]` helper, and then using that here + in the KaTex module.

PR #126, opened 2022-08-18
🐋 → ❄️

ncfavier:
Changes the build workflow to use Nix instead of a Docker container, and refactors things a bit. The most important change is that `default.nix` now takes an `interactive` argument which, when enabled (the default), adds the full `ghcWithPackages` to the environment (which includes Agda binaries). This is disabled in CI and in the Docker image.

The non-Nix build methods still work.

When everything is cached, a [trivial run](https://github.com/ncfavier/1lab/runs/7895145982?check_suite_focus=true) finishes in 2m30s and fetches about 300 MB of data, which seems acceptable.

To get caching working, you (Amy) will need to create a (free) binary cache at https://www.cachix.org/ (this PR assumes it's named `1lab`), select "API tokens" for the access method, generate an authentication token and add it as an action secret named `CACHIX_AUTH_TOKEN`.

To *use* the cache, Nix users can run `cachix use 1lab` if they have Cachix installed, or add the cache to `substituters` and `trusted-public-keys` manually in their Nix config. No more building Agda!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,7 +3,6 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
-    container: pltamy/1lab:latest
 
     env:
       mailmap: ${{ secrets.MAILMAP }}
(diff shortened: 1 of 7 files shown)


SquidDev:
How does this work for PRs? CACHIX_AUTH_TOKEN won't be set - does the cache still work, it's just read-only?

ncfavier:
[Yes.](https://github.com/ncfavier/1lab/runs/7897609260?check_suite_focus=true)

SquidDev:
Ahh lovely, that's perfect. Thank you!

plt-amy:
The two of you _need_ to add yourselves to the authors page. Thanks so much @ncfavier! Cool as heck

PR #130, opened 2022-08-25
chore: make Pandoc shut up about missing translations

ncfavier:
Sets the default translation language to English to avoid warnings like "The term Abstract has no translation defined."

And removes references before compressing the executable, otherwise hashes can [escape](https://github.com/NixOS/nixpkgs/pull/188299).


plt-amy:
Bless your soul

PR #131, opened 2022-08-30
Discrete Fibrations and Presheaves

TOTBWF:
# Description

This is a little one: it defines the displayed category of elements, and also constructs both halves of the equivalence between discrete fibrations and presheaves (though it doesn't show they are an equivalence for reasons explained below)

## Notes
I considered redefining the non-displayed category of elements in terms of the total category of the displayed one, but I figured we might want to discuss that one first 😛. Also, it would be rename `Discrete-fibration` to `is-discrete-fibration`, and then define the bundled form. This would make showing the equivalence between presheaves and discrete fibrations a bit less tedious, as we don't need to deal with uncurrying some things.

### DIFF
--- a/src/Cat/Displayed/Cartesian/Discrete.lagda.md
+++ b/src/Cat/Displayed/Cartesian/Discrete.lagda.md
@@ -1,7 +1,13 @@
+---
+description: |
+  We define discrete fibrations,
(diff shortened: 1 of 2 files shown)


plt-amy (approved):
Cool! :smile:

PR #142, opened 2022-10-17
Binary Products Solver

TOTBWF:
# Description

This patch adds a literate solver for categories with all binary products. Hopefully this should make life a bit easier!

### DIFF
--- a/src/Cat/Diagram/Product.lagda.md
+++ b/src/Cat/Diagram/Product.lagda.md
@@ -253,4 +253,14 @@ separate top-level definitions:
   ⟨⟩∘ : ∀ {Q R} {p1 : Hom Q a} {p2 : Hom Q b} (f : Hom R Q)
       → ⟨ p1 , p2 ⟩ ∘ f ≡ ⟨ p1 ∘ f , p2 ∘ f ⟩
   ⟨⟩∘ f = is-product.⟨⟩∘ (hasprods _ _ .has-is-product) f
+
+  ⟨⟩-unique : ∀ {Q} {p1 : Hom Q A} {p2 : Hom Q B}
+             → (other : Hom Q (A ⊗ B))
+             → π₁ ∘ other ≡ p1
+             → π₂ ∘ other ≡ p2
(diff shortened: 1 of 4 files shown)


TOTBWF:
Ugh, doesn't quite work for `Cat.Monoidal.Instances.Cartesian`, due to some annoying reflection edge cases where agda decides to expand out `Cartesian.⟨_,_⟩` before it hits reflection.

TOTBWF:
So the solver is able to handle the associator proofs in `Cat.Monoidal.Instances.Cartesian`, but you need to explicitly write out the goals using equational reasoning. If you try to do something like
```agda
      ni .eta∘inv x = products! C prods
```

then we get the following failure:
```
Could not equate the following expressions:
  (_445 prods term Precategory.∘ .eta ni x) (.inv ni x)
And
  .Precategory.id (_445 prods term)
Reflected expressions
  NbE.‶ _471 prods term x ‶ NbE.‶∘‶ NbE.‶ _495 prods term x ‶
And
  NbE.‶id‶
Computed normal forms
  (C Precategory.∘ _528 prods term x)
  ((C Precategory.∘ _532 prods term x) (.Precategory.id C))
And
  .Precategory.id C
when checking that the expression unquote (products! C prods) has
type
(_D_445 Precategory.∘ ni .eta x) (ni .inv x) ≡
Precategory.id _D_445
```

It seems like Agda is losing track of some things, which is really weird! If we just do
```agda
      ni .eta∘inv x = ?
```

we get the following goal:
```
Goal: ⟨ π₁ ∘ π₁ , ⟨ π₂ ∘ π₁ , π₂ ⟩ ⟩ ∘
      ⟨ ⟨ π₁ , π₁ ∘ π₂ ⟩ , π₂ ∘ π₂ ⟩
      ≡ id
```

I'm not sure what's going on here...

plt-amy:
having to write out the goal but not the proof is already a massive improvement if you ask me. thanks!

plt-amy (approved):

PR #159, opened 2022-12-07
defn: refactor F∘-functor, global Kan extensions, uniqueness of Lan

plt-amy:
just getting a quick CI run :slightly_smiling_face:


plt-amy:
"just a quick CI run", they said

PR #160, opened 2022-12-09
Coends in Set

TOTBWF:
# Description

This lil PR explicitly constructs coends in Sets as a coequaliser. This is needed for some of the Functor Kit experiments I'm running, but it's standalone enough to factor out.


plt-amy:
Just so I can be sure, is the point here just that this computes better than `cocomplete→coend`, so actually working with the concrete coend becomes feasible?

TOTBWF:
Precisely. We could generalize this to use arbitrary coequalisers ofc, but this is meant to be optimized for `Sets`. If this didn't involve a functor I would have put in `Data.Set`.

PR #161, opened 2022-12-13
Tiny Shakefile changes

SquidDev:
- Use `git shortlog` to compute authors/co-authors. This means we only need to run one git command for each file, rather than two.
 - Fix "Cannot find Some.Agda.Module" errors.


plt-amy:
> Fix "Cannot find Some.Agda.Module" errors.

i owe you my life

SquidDev:
I don't think this fixes the recursive dependency issues which happen occasionally. I've not been able to reproduce them again, but sure it'll happen at some point.

plt-amy:
I think it happens when you have an error in a module `A` depending on a module `B`, then you change `B`

PR #169, opened 2023-01-09
Lower minimum screen width for desktop

marshall-lee:
On my laptop, I learned that sidebar even exists only when I suddenly zoomed out the page. Before that, I just wondered why the website is so minimalistic.

AFAIK 1280px is quite a popular choice for determining desktop screen. 13.3-inch macbook has 1280px for instance.

### DIFF
--- a/support/web/css/default.scss
+++ b/support/web/css/default.scss
@@ -258,7 +258,7 @@ table {
   }
 }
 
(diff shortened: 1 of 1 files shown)


plt-amy:
Thanks for your contribution!

PR #183, opened 2023-01-29
chore: throw on KaTeX errors

ncfavier:
Note that since https://github.com/plt-amy/1lab/pull/173 was squash merged and the commit messages discarded, I'm now listed as an author on a bunch of pages I didn't meaningfully contribute to 😩

Ideally we should have a list of revisions to ignore, à la `git blame --ignore-revs-file`, but shortlog doesn't seem to support anything like that and git blame can't output the information we want.


plt-amy:
You know I was sure I had gotten all of them... I even went grepping for katex-error in the HTML output. Thanks 🙃

PR #193, opened 2023-02-13
Adjoints have Kan extensions, revisited

ncfavier:
An adjunction $L ⊣ R$ induces an adjunction $- ∘ R ⊣ - ∘ L$ between precomposition functors, which means that

- $L$ has left Kan extensions given by precomposition with $R$
- $R$ has right Kan extensions given by precomposition with $L$

(There's also a "dual duality" with Kan *lifts* and postcomposition, but we don't have those yet.)

This is an IMO more intuitive rephrasing of the fact that [adjoints are **absolute** Kan extensions](https://ncatlab.org/nlab/show/adjoint+functor#in_terms_of_kan_extensionsliftings), but I couldn't find it anywhere (didn't look very hard).

**Notation issue**: from what I can tell, $f^\*$ usually denotes morphisms that are "flipped" with respect to $f$ whereas $f_!$ and $f_\*$ denote morphisms in the same direction. For example, the nlab page on [Kan extensions](https://ncatlab.org/nlab/show/Kan+extension#OrdinaryKanExtensions) uses $p^\*$ for the precomposition functor $[C', D] \to [C, D]$ and $p_!$ for its left adjoint. The 1lab seems to do the [opposite](https://1lab.dev/Cat.Instances.Functor.Compose.html#functoriality-of-functor-composition), which is a bit confusing. I think we should at least swap them, or just get rid of this confusing convention and call them `_∘-` and `-∘_`.

## Checklist

- [ ] The imports are sorted (use `find -type f -name \*.agda -or -name \*.lagda.md | xargs support/sort-imports.hs`)
  (They weren't before)

- [X] All code blocks have "agda" as their language. This is so that

- [X] Proofs are explained to a satisfactory degree; This is subjective,

### DIFF
--- a/src/Cat/Functor/Adjoint.lagda.md
+++ b/src/Cat/Functor/Adjoint.lagda.md
@@ -667,7 +667,7 @@ between [postcomposition and precomposition functors], respectively:
 ```agda
   open import Cat.Instances.Functor.Compose
 
-  postcomposite-adjunction : (L ^*) {D = E} ⊣ R ^*
+  postcomposite-adjunction : postcompose L {D = E} ⊣ postcompose R
   postcomposite-adjunction .unit .η F .η = (adj.unit ◂ F) .η
   postcomposite-adjunction .unit .η F .is-natural = (adj.unit ◂ F) .is-natural
   postcomposite-adjunction .unit .is-natural F G α = Nat-path λ _ → adj.unit.is-natural _ _ _
@@ -677,7 +677,7 @@ between [postcomposition and precomposition functors], respectively:
   postcomposite-adjunction .zig = Nat-path λ _ → adj.zig
   postcomposite-adjunction .zag = Nat-path λ _ → adj.zag
 
-  precomposite-adjunction : (R !) {D = E} ⊣ L !
+  precomposite-adjunction : precompose R {D = E} ⊣ precompose L
   precomposite-adjunction .unit .η F .η = (F ▸ adj.unit) .η
   precomposite-adjunction .unit .η F .is-natural = (F ▸ adj.unit) .is-natural
   precomposite-adjunction .unit .is-natural F G α = Nat-path λ _ → sym (α .is-natural _ _ _)
--- a/src/Cat/Functor/Adjoint.lagda.md
+++ b/src/Cat/Functor/Adjoint.lagda.md
@@ -666,23 +666,20 @@ between [postcomposition and precomposition functors], respectively:
 
 ```agda
   open import Cat.Instances.Functor.Compose
+  open import Cat.Functor.Coherence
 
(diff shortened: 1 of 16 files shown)


plt-amy:
> The 1lab seems to do the [opposite](https://1lab.dev/Cat.Instances.Functor.Compose.html#functoriality-of-functor-composition), which is a bit confusing. I think we should at least swap them, or just get rid of this confusing convention and call them _∘- and -∘_.

Agreed but I don't like the symbolic names (for what it's worth, I regret going with `_!` and `_^*`, too). We can call them `postcompose` and `precompose` as you've done for the new lemmas.

src/Cat/Functor/Adjoint.lagda.md
```diff
@@ -594,7 +596,37 @@ module _ {L : Functor C D} {R : Functor D C} (adj : L ⊣ R) where
   opposite-adjunction .counit .is-natural x y f = sym (adj.unit.is-natural _ _ _)
   opposite-adjunction .zig = adj.zag
   opposite-adjunction .zag = adj.zig
+```
 
+As well as adjunctions $L \circ - \dashv R \circ -$ and $- \circ R \dashv - \circ L$
+between [postcomposition and precomposition functors](Cat.Instances.Functor.Compose.html), respectively:
```
plt-amy, line 602:
```suggestion
between [postcomposition and precomposition functors] respectively:

[postcomposition and precomposition]: Cat.Instances.Functor.Compose.html
```

src/Cat/Functor/Kan/Global.lagda.md
```diff
@@ -113,3 +113,14 @@ adjoint→Lan F F⊣p! G = ext where
   ext .σ-uniq x = Equiv.injective (_ , L-adjunct-is-equiv F⊣p!)
     (L-R-adjunct F⊣p! _ ∙ x)
 ```
+
+In particular, if $p$ itself has a *right* adjoint $p \dashv r$, then left Kan
+extensions along $p$ are given by `precomposition`{.Agda ident=!} with $r$:
```
plt-amy, line 118:
```suggestion
extensions along $p$ are given by `precomposition`{.Agda ident=_!} with $r$:
```

ncfavier:
> I don't like the symbolic names

Why not? They have the advantage that you don't have to keep mentally translating between $H_!$ (or $H^*$, or $- \circ H$) and `precompose H` when reading a page like https://1lab.dev/Cat.Univalent.Rezk.Universal.html. I think the cohesiveness there is nice.

ncfavier:
I've renamed `precompose` and `postcompose` as requested and rewritten Cat.Functor.Kan.Adjoint completely.

Issue #91, opened 2022-05-17
Lex functors preserve monos


plt-amy:
Should be a super simple theorem to write up. We already have a characterisation of monos as certain squares being pullbacks.
Bonus points: Rex functors preserve epis, after #90 is done.

ncfavier:
Closed by https://github.com/plt-amy/1lab/pull/196?

PR #196, opened 2023-03-06
lex/rex preserve monos/epis and pushout properties

Seiryn21:
I hope this is what you envisioned in #91 , also I'm still a beginner in category theory so I hope the textual parts are fine

### DIFF
--- a/src/Cat/Diagram/Pushout/Properties.lagda.md
+++ b/src/Cat/Diagram/Pushout/Properties.lagda.md
@@ -1,0 +1,72 @@
+```agda
+open import Cat.Prelude
+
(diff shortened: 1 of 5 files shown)


src/Cat/Diagram/Pushout/Properties.lagda.md
```diff
@@ -0,0 +1,72 @@
...
+```
+
+<!--
+```agda
+open Cat.Diagram.Pushout C
+open Cat.Reasoning C
+open is-pushout
+
+```
+--!>
```
plt-amy, line 17: Just `-->`

plt-amy:
Thank you!

PR #202, opened 2023-04-02
Use rem for managing layout

SquidDev:
# Description
While `ch` and `em` are conceptually much nicer, the fact that they rely on the current element's font size makes them less useful in managing layout as the units are not the same across the page. This causes a couple of problems:

 - @media breakpoints are entirely incorrect.
 - The sidebar and main contents have different padding, which makes them look slightly wonky.

Instead we use `rem` universally (well, in the places that it matters). This does slightly adjust the layout of the page, but it should now be better behaved.

### DIFF
--- a/support/web/css/default.scss
+++ b/support/web/css/default.scss
@@ -69,12 +69,12 @@ body {
 }
 
 main {
-  max-width: 85ch;
+  max-width: 80rem;
   width: 100%;
   margin: 0px auto 0px auto;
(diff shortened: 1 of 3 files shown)


SquidDev:
### Before
A screenshot of the 1lab.dev homepage.

### After
A screenshot of my local copy of the 1lab.dev homepage. The layout is slightly different: for instance, the main content is a little narrower.

PR #206, opened 2023-04-19
Misc fixes

ncfavier:


.github/workflows/build.yml
```diff
@@ -14,7 +14,7 @@ jobs:
           fetch-depth: 0 # we need the commit history for authors
 
       - name: Install Nix ❄️
-        uses: cachix/install-nix-action@v18
+        uses: cachix/install-nix-action@v20
```
ncfavier, line 17: Fixes https://github.com/plt-amy/1lab/actions/runs/4742421368/jobs/8420675573?pr=205#step:4:11 (not sure why this only seems to affect forks).

src/1Lab/Reflection/Subst.agda
```diff
@@ -42,7 +42,7 @@ singletonS n u = map (λ i → var i []) (count (n - 1)) ++# u ∷ (raiseS n)
   where
     count : Nat → List Nat
     count zero = []
-    count (suc n) = 1 ∷ map suc (count n)
+    count (suc n) = 0 ∷ map suc (count n)
```
ncfavier, line 45: https://github.com/agda/agda/blob/2c05d884cdbbaa3ddcde901faea216d2a4859cfc/src/full/Agda/TypeChecking/Substitute/Class.hs#L136

PR #211, opened 2023-04-22
Generalise regularity tactic

ncfavier:
Inspired by a [question](https://agda.zulipchat.com/#narrow/stream/260790-cubical/topic/How.20to.20prove.20equality.20of.20subst'ed.20dependent.20functions.3F/near/351670006) on the Agda Zulip.

Remove the restriction that `φ = i0`, so that the tactic works on general `transp` applications.

This allows proving a characterisation of paths between Poly maps, which involves a nasty goal with nested `transp`s that the tactic previously couldn't solve.

### DIFF
--- a/src/1Lab/Reflection/Regularity.agda
+++ b/src/1Lab/Reflection/Regularity.agda
@@ -11,26 +11,22 @@ module 1Lab.Reflection.Regularity where
 A tactic for reducing "transport refl x" in other-wise normal terms. The
 implementation is actually surprisingly simple: A term of the form (e.g.)
 
-    transport refl (f (transport refl x))
+    transp (λ _ → A) i0 (f (transp (λ _ → B) i0 x))
 
 is already a blueprint for how to normalise it. We simply have to turn it into
 
-    λ i → transport-refl (f (transport-refl x i)) i
+    λ i → transp (λ _ → A) i (f (transp (λ _ → B) i x))
 
-The implementation presents itself: Raise the original term, and look
-through to find reflexive transports. These can be replaced by an
-application of transport-refl.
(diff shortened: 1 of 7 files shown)


plt-amy (approved):

src/1Lab/Reflection/Regularity.agda
```diff
@@ -11,26 +11,22 @@ module 1Lab.Reflection.Regularity where
...
+recursively (inside-out), on the condition that Al is constant when i = i1.
 -}
 
 private
-  -- Ahem, not transport-refl, but a tiny tiny wrapper around it. The
-  -- reason we have 'regular' is to mark terms we're *leaving*: The
-  -- translation proceeds inside out, so to have a way to stop, we mark
-  -- terms we have already visited entirely using `regular`.
-  regular : ∀ {ℓ} {A : Type ℓ} (x : A) → transport refl x ≡ x
-  regular x = transport-refl x
```
ncfavier, line 31: I don't think we were actually using this as a marker?

src/1Lab/Reflection/Regularity.agda
```diff
@@ -159,9 +155,7 @@ private
 module Regularity where
   open Regularity-precision public
   -- The reflection interface: Regularity.reduce! will, well, reduce a
-  -- term. The tactic is robust enough to grow terms, if you invert the
-  -- path, as well. There's a lot of blocking involved in making this
-  -- work.
```
ncfavier, line 164: It's not, but `precise! refl` does that.

ncfavier:
I've [slain](https://github.com/plt-amy/1lab/pull/211/commits/359d878f622547abef096929c701add0a2eca3c7) a couple of `transp` monsters, cc @TOTBWF

PR #225, opened 2023-05-06
Pseudomonic Functors, Wide Subcategories, Core of a Category, Fibration of Objects

TOTBWF:
# Description

This is a bit of a grab bag of results required to give the fibration of objects a proper treatment. 
It defines:
- The fibration of objects, which is the fibred version of the core of a category.
- The core of a category
- Wide subcategories
- Pseudomonic functors

The core is needed to explain the fibration of objects, but defining it is best done via wide subcategories, which in turn need pseudomonic functors to properly explain!

## Notes

I've omitted the requirement that a wide subcategory contain all isos, as it's not *strictly* needed to construct. However, this is required for the inclusion to be pseudomonic.


src/Cat/Functor/Base.lagda.md
```diff
@@ -247,3 +259,86 @@ module _ {C : Precategory o h} {D : Precategory o₁ h₁} where
...
+      faithful : is-faithful F
+      isos-full : is-full-on-isos F
+
+  open is-pseudomonic
+```
+
+Somewhat surprisingly, pseudomonic functors are [conservative]. To
+see this, note that invertability is a property of morphisms, so we
+can eliminate out of the truncated isomorphism we obtain via
+pseudomonadicity. Once we have our hands on the isomorphism, we
```
ncfavier, line 295: I don't think "pseudomonadicity" works, but I also have no idea what the noun should be. "Pseudomonicity"?

TOTBWF: Total typo lol, had 2-categories on the brain 😅

PR #227, opened 2023-05-12
Generic Objects

TOTBWF:
# Description

This PR defines generic objects, and proves some properties about them. Most notably, it proves a handful of the results from https://arxiv.org/abs/2210.04202.

### DIFF
--- a/src/1Lab/Path.lagda.md
+++ b/src/1Lab/Path.lagda.md
@@ -1608,16 +1608,6 @@ funext : ∀ {a b} {A : Type a} {B : A → Type b}
 funext p i x = p x i
 ```
 
-<!--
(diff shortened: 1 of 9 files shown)


src/1Lab/Path.lagda.md
```diff
@@ -1608,6 +1608,16 @@ funext : ∀ {a b} {A : Type a} {B : A → Type b}
 funext p i x = p x i
 ```
 
+<!--
+```agda
+funextp : ∀ {a b} {A : Type a} {B : A → I → Type b}
```
plt-amy, line 1613: https://1lab.dev/1Lab.Type.Pi.html#3191

PR #228, opened 2023-05-13
Add cachebusters to CSS and JS

SquidDev:
Just adds a `?v=filehash` to the CSS and JS files. This does not apply to Agda-only files (so those using the "code" template), but I think that's probably fine for now.

~~This does massively increase build times when changing CSS. It looks like some markdown files are taking >5 seconds to be converted. I'm not sure why this is, want to do some profiling before merging.~~

### DIFF
--- a/support/shake/1lab-shake.cabal
+++ b/support/shake/1lab-shake.cabal
@@ -77,6 +77,7 @@ executable shake
       HTML.Emit
     , Shake.AgdaCompile
     , Shake.Diagram
+    , Shake.Digest
     , Shake.Git
     , Shake.KaTeX
     , Shake.LinkGraph
--- a/support/shake/app/Main.hs
+++ b/support/shake/app/Main.hs
@@ -33,6 +33,7 @@ import Shake.LinkGraph
 import Shake.Markdown
 import Shake.Modules
 import Shake.Diagram
+import Shake.Digest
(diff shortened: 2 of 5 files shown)


SquidDev:
Adding a couple of `timedM` calls, shows we're spending a lot more time rendering Markdown than I'd expect.

|                         Label |   Time (s)|
|------------------------------ | ----------|
|                  Check markup |       0.71|
|                    Fix markup |       4.12|
|                Parse Markdown |       4.25|
|                Patch markdown |       0.46|
|               Render Markdown |      21.25|
|                        Search |       0.29|
|                    Write file |       2.77|
|                   patchInline |       0.46|
|                  patchInlines |       0.36|
|                 uncommentAgda |       0.32|

Not sure why this is - I'd expect Pandoc to be relatively sensible here. Might try to see if we can build a profiling build of 1lab.

SquidDev:
After some profiling, most of the time was being spent inside Skylighting (I was running with `--skip-agda`). I've patched Skylighting to pre-compute the regexes, ensuring they're cached between code blocks, which brings the total build time down to ~7s (from ~15s) - not ideal, but I think workable now.

Not 100% wild about the maintainability of this - I will submit this patch to upstream, so we may want to wait if/until that is merged.

PR #237, opened 2023-06-18
fix: Make category reflection code  more precise

TOTBWF:
# Description

Previously, we were doing category reflection by blocking reduction of *all* category fields. This causes problems for things like Kleisli categories; blocking reduction of all compositions means that we can't unfold composition in the kleisli category, which in turn means that the monad solver fails.

The simplest way to fix this is to leverage unification to determine when we encounter field projections for a *specific* category, which I've done here.

### DIFF
--- a/src/1Lab/Reflection.lagda.md
+++ b/src/1Lab/Reflection.lagda.md
@@ -549,6 +549,10 @@ _ term=? _ = false
 “refl” : Term
 “refl” = def (quote refl) []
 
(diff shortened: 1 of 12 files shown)


TOTBWF:
It looks like this causes a pretty serious performance regression, so let's shelve this for now ☹️

TOTBWF:
Closing for now due to the perf impacts.

PR #239, opened 2023-07-03
HLevel automation fix

cmcmA20:
@plt-amy Thank you for the effort you put in this project. I do not have the time to revamp the automation like I did in my library, but I hope you'll find this fix useful.


cmcmA20:
You can also see it in the full generality here:
https://github.com/cmcmA20/cubical-mini/blob/master/src/Meta/Search/Base.agda

Please upgrade your automation. I could just use yours and deal with my domain of interest, but bugs were too annoying. But thank you for motivating me to finally learn some agda metaprogramming. :-)

PR #240, opened 2023-07-17
Equivalence facts

TOTBWF:
# Description

Some misc. properties of various classes of morphisms, along with some properties of equivalences.

### DIFF
--- a/src/1Lab/Equiv.lagda.md
+++ b/src/1Lab/Equiv.lagda.md
@@ -714,19 +714,6 @@ lift-inj
   → a ≡ b
 lift-inj p = ap Lift.lower p
 
-Lift-pathp
-  : ∀ {ℓ ℓ'} {A B : Type ℓ} {a : Lift ℓ' A} {b : Lift ℓ' B}
-  → (p : A ≡ B)
-  → PathP (λ i → p i) (Lift.lower a) (Lift.lower b)
-  → PathP (λ i → Lift ℓ' (p i)) a b
-Lift-pathp p q i .Lift.lower = q i
-
-Lift-path
-  : ∀ {ℓ ℓ'} {A : Type ℓ} {a a' : Lift ℓ' A}
-  → Lift.lower a ≡ Lift.lower a'
-  → a ≡ a'
-Lift-path = Lift-pathp refl
-
 Lift-≃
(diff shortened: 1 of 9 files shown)


src/1Lab/Equiv.lagda.md
```diff
@@ -707,5 +707,37 @@ prop-ext pprop qprop p→q q→p .snd .is-eqv y .paths (p' , path) =
...
+  → (p : A ≡ B)
+  → PathP (λ i → p i) (Lift.lower a) (Lift.lower b)
+  → PathP (λ i → Lift ℓ' (p i)) a b
+Lift-pathp p q i .Lift.lower = q i
+
+Lift-path
+  : ∀ {ℓ ℓ'} {A : Type ℓ} {a a' : Lift ℓ' A}
+  → Lift.lower a ≡ Lift.lower a'
+  → a ≡ a'
+Lift-path = Lift-pathp refl
```
plt-amy, line 728: Is there any case where this would be better than `ap lift`?

src/Cat/Functor/Base.lagda.md
```diff
@@ -346,3 +347,23 @@ essentially injective.
          D.≅-pathp refl refl (equiv→counit ff (D.to f)))
 ```
 
+## Equivalence on Objects Functors
+
+A functor $F : \cC \to \cD$ is an equivalence on objects if it's action
```
plt-amy, line 352: its

src/Cat/Functor/Base.lagda.md
```diff
@@ -346,3 +347,23 @@ essentially injective.
...
+on objects is an equivalence (shocking!).
```
plt-amy, line 353: drop the sarcasm

src/Cat/Functor/Base.lagda.md
plt-amy: also "equivalence on objects" should be bold

plt-amy (changes requested):

src/1Lab/Equiv.lagda.md
TOTBWF: I think I was on autopilot here, whoops!

plt-amy (approved):

PR #241, opened 2023-07-25
Comprehension coproducts

TOTBWF:
# Description

This PR defines coproducts with respect to a comprehension category, and also defines strong and very strong coproducts.
It depends on #232, and should get rebased before merging.


TOTBWF:
#232 has been merged, so this is good to go!

PR #244, opened 2023-07-28
Explode Cat.Instances.Functor

plt-amy:
Disappear. This module is an eyesore.


src/Cat/Functor/Closed.lagda.md
TOTBWF, line 1: Maybe leave a TODO here for prose

TOTBWF (dismissed):
Looks great! Only one minor nit: we should add a TODO for prose for Functor.Closed

PR #255, opened 2023-09-12
A lot of minor cleanup

plt-amy:
How's that saying go, it's always spring somewhere?


CONTRIBUTING.md
```diff
@@ -9,63 +9,191 @@ guidelines, and not strict rules: If you're unsure of something, hop on
...
+
+There is a dedicated operator for projecting the "underlying type" of
+anything for which this concept makes sense: `⌞ T ⌟`; see
+[1Lab.Underlying](https://1lab.dev/1Lab.Underlying.html). Underlying
+instances are available for:
+
+- `Type` (no effect)
+- `n-Type` and `Ω` (projects the underlying type)
+- A type like `Σ U ...` (projects the underlying type of the first component)
+- Concrete groups (projects the delooping)
```
ncfavier, line 143: Maybe keep the list open because there's no way we're going to remember to update it.

default.nix
```diff
@@ -8,12 +8,12 @@ let
...
   inherit (pkgs) lib;
 
-  our-ghc = pkgs.labHaskellPackages.ghcWithPackages (ps: with ps; [
+  our-ghc = pkgs.labHaskellPackages.ghcWithPackages (ps: with ps; ([
     shake directory tagsoup
     text containers uri-encode
     process aeson Agda pandoc SHA
     fsnotify
-  ]);
+  ] ++ (if inNixShell then [ haskell-language-server ] else [])));
```
ncfavier, line 16:
```suggestion
  ] ++ (if interactive then [ haskell-language-server ] else [])));
```

We don't want this in CI

support/sort-imports.hs
```diff
@@ -17,10 +17,11 @@ import Data.Function (on)
 import Data.Foldable hiding (find)
 import Data.Ord
 
+import Development.Shake.FilePath
+import Development.Shake
```
ncfavier, line 21: The dependencies in the shebang comment need updating

default.nix
ncfavier: Also this is a mass rebuild for me (144 Haskell packages...). I wonder if we could use a version of nixpkgs/GHC with a binary cache available (if not, I guess the 1lab's cachix will have to do).

cabal.project
```diff
@@ -1,6 +1,6 @@
 packages: support/shake/1lab-shake.cabal
 
 source-repository-package
-    type: git
-    location: https://github.com/agda/agda.git
-    tag: 269204b682d87778ba1038193db05e728cfadbdb
+  type: git
+  location: https://github.com/agda/agda.git
+  tag: 269204b682d87778ba1038193db05e728cfadbdb
```
ncfavier, line 6: This stanza actually makes cabal rebuild Agda instead of using the version provided by Nix, and this is not just because the revision is out of sync with `support/nix/dep/Agda`.

I don't know whether we can tell cabal to use whatever `Agda` is installed locally and fall back to git otherwise.

plt-amy: Why are you using nix *and* cabal?

plt-amy: Why are you using nix *and* cabal?

ncfavier: HLS uses cabal

plt-amy: behold, hie.yaml:
```yaml
cradle:
  multi:
    - path: "./src"
      config:
        cradle:
          none:
    - path: "./support/shake"
      config:
        cradle:
          bios:
            shell: "echo '-isupport/shake/app' > $HIE_BIOS_OUTPUT; find support/shake/app -type f -name '*.hs' -or -name '*.hs-boot' >> $HIE_BIOS_OUTPUT"
```

ncfavier: Interesting, that works if I launch `haskell-language-server` from the terminal, but when codium does it it spawns a process tree that looks like it's fetching agda-stdlib from github, but takes an eternity.

I give up for now, I'll just delete cabal.project locally whenever I need to work on the Haskell files.

ncfavier: Until I switch branches, apparently 😩

CONTRIBUTING.md
```diff
@@ -9,63 +9,192 @@ guidelines, and not strict rules: If you're unsure of something, hop on
...
+  [the definition]
+  ```
+
+Wikilink targets are case insensitive, and get "mangled" for whitespace:
+the anchor `least-upper-bound` can be referred to as `[[least upper
+bound]]`, or `[[lub|least upper bound]]` if different textual content is
+required, or `[[lub|least-upper-bound]]`. White space **should** be used
+instead of dashes at use sites.
+
+**Warning**: A wikilink can not span multiple lines. If a wikilink would
```
ncfavier, line 54: Wasn't the point of `mangleMarkdown` to allow this?

plt-amy: oh yeah

CONTRIBUTING.md
```diff
@@ -9,63 +9,229 @@ guidelines, and not strict rules: If you're unsure of something, hop on
...
+be referred to in prose using `` `name`{.Agda} `` (an inline code block
+with class `Agda`). The [HoTT book comparison] uses this feature
+extensively.
+
+[HoTT book comparison]: https://github.com/plt-amy/1lab/blob/main/src/HoTT.lagda.md
+
+If the definition does not naturally appear in the code, the following
+idiom can be used. The implementation of identifier links prevents their
+use in a mutually-recursive way, since it should only be used for
+actualy. Cyclic links **should** be [wikilinks](#links) instead.
```
ncfavier, line 66: Used for what now? 😅

PR #256, opened 2023-09-12
wip: workflow for previewing the website

plt-amy:
Only the prose, though, to get an idea of how things render. Time to write YAML...


ncfavier:
> Only the prose

We are building the whole thing anyway, why not do the previewing in the main build action?

plt-amy:
@ncfavier The prose takes ~5 minutes to build clean, the full website takes almost half an hour

PR #260, opened 2023-09-14
web: search for modules and definitions

ncfavier:
Allows searching for modules by name and for definitions.

Extract search data before rendering math to HTML so that the plain writer can render it in descriptions.

Clear the default base URL so that local files are used in development.


ncfavier:
Needs a cache bust probably

ncfavier:
oh yeah oepsie

PR #261, opened 2023-09-14
CSS fixes

ncfavier:
Prevent the main body from flowing out of the screen, and reduce the font size a bit.

I tweaked the sizes based on what looks good on my screens but feel free to adjust further.


support/web/css/default.scss
```diff
@@ -265,7 +261,9 @@ table {
 
     div#post-toc-container {
       display: grid;
-      grid-template-columns: 0.5fr 2fr 0.5fr;
+
+      // The 0 minimum is required to avoid overflow, see https://css-tricks.com/preventing-a-grid-blowout/
+      grid-template-columns: 1fr minmax(0, 4fr);
```
plt-amy, line 266: There's a reason the body was centered:

PR #267, opened 2023-09-28
chore: synchronise HTML backend

ncfavier:
https://github.com/agda/agda/pull/6880

It would be nice to upstream whatever changes we have to the HTML backend so that we can stay in sync.


SquidDev:
> It would be nice to upstream whatever changes we have to the HTML backend so that we can stay in sync.

Yeah, this has been on my todo list for ages (https://github.com/agda/agda/issues/6311). I'm not 100% sure how we want to handle the generated type JSON here, as that possibly makes less sense upstream.

PR #272, opened 2023-10-25
Extensionality tactic

plt-amy:
Implements a metaprogram `ext` which finds a "good" identity system for a given type, defaulting to `Path` at leaves. To implement default instances, the search is split between the `Extensional` class that actually carries the data and a loop-breaker `Extensionality` class that's used to match on the goal.

There's also a helper `trivialᵉ` for the most common case (surprisingly) where the values are extensionally equal, i.e. things like `funext λ x → Nat-path λ y → refl`.

Currently restricted to non-dependent types, so e.g. `Nat → B G →∙ B H` will stop at `∀ x → Path (B G →∙ B H) _ _` since nothing can match the `Σ`. Reductions are implemented for functions, pairs, natural transformations, and most kinds of morphism in concrete categories. `Homomorphism-path` can be consistently replaced by `ext`, and the search is robust enough to not need any new type annotations basically anywhere.

### DIFF
--- a/src/1Lab/Extensionality.agda
+++ b/src/1Lab/Extensionality.agda
@@ -241,7 +241,7 @@ private
       `x ← quoteTC x
       `y ← quoteTC y
       typeError
-        [ "trivialᵉ failed: the values\n  "
+        [ "trivial! failed: the values\n  "
         , termErr `x
         , "\nand\n  "
         , termErr `y
@@ -249,7 +249,7 @@ private
         ]
 
 {-
-trivialᵉ serves to replace proofs like
+trivial! serves to replace proofs like
 
   Nat-path λ x → funext λ y → Nat-path λ z → Homomorphism-path λ a → refl
 
@@ -263,12 +263,12 @@ up.
 -}
 
 opaque
-  trivialᵉ
+  trivial!
     : ∀ {ℓ ℓr} {A : Type ℓ} {x y : A}
     → ⦃ r : Extensional A ℓr ⦄
     → {@(tactic trivial-worker r x y) p : Pathᵉ r x y}
     → x ≡ y
-  trivialᵉ ⦃ r ⦄ {p = p} = r .idsᵉ .to-path p
+  trivial! ⦃ r ⦄ {p = p} = r .idsᵉ .to-path p
 
 Pathᵉ-is-hlevel
   : ∀ {ℓ ℓr} {A : Type ℓ} n (sa : Extensional A ℓr)
--- a/src/1Lab/Extensionality.agda
+++ b/src/1Lab/Extensionality.agda
@@ -85,7 +85,7 @@ extensional A goal = do
   unify goal =<< find-extensionality `A
 
 {-
-Unlike extensionalᶠ, which is parametrised by a type, extensionalᶠ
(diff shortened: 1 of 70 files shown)


src/1Lab/Extensionality.agda
```diff
@@ -0,0 +1,299 @@
...
+
+-- Entry point for getting hold of an 'Extensional' instance:
+extensional : ∀ {ℓ} (A : Type ℓ) → Term → TC ⊤
+extensional A goal = do
+  `A ← quoteTC A
+  checkType goal (def (quote Extensional) [ argN `A , argN unknown ])
+  unify goal =<< find-extensionality `A
+
+{-
+Unlike extensionalᶠ, which is parametrised by a type, extensionalᶠ
```
ncfavier, line 88:
```suggestion
Unlike extensional, which is parametrised by a type, extensionalᶠ
```

ncfavier (dismissed):

src/1Lab/Extensionality.agda
```diff
@@ -0,0 +1,299 @@
...
+
+  ext λ x y z a → refl
+
+and that argument is precisely reflexivity for the computed identity
+system which 'ext' is using. By the way, this example is totally made
+up.
+-}
+
+opaque
+  trivialᵉ
```
TOTBWF, line 266: Maybe call this `trivial!`?

src/Cat/Base.lagda.md
```diff
@@ -575,3 +577,35 @@ is a proposition:
...
+Extensional-natural-transformation {sa = sa} .reflᵉ x i = reflᵉ (sa i) (x .η i)
+Extensional-natural-transformation {sa = sa} .idsᵉ .to-path x = Nat-path λ i →
+  sa _ .idsᵉ .to-path (x i)
+Extensional-natural-transformation {D = D} {sa = sa} .idsᵉ .to-path-over h =
+  is-prop→pathp
+    (λ i → Π-is-hlevel 1
+      (λ _ → is-hlevel≃ 1 (identity-system-gives-path (sa _ .idsᵉ)) (D .Hom-set _ _ _ _)))
+    _ _
+
+instance
```
TOTBWF, line 603: A note on why we use `Extensionality` here would be nice; I expect that this will be a lot of folks entrypoint into the internals of the tactic.

src/Cat/Morphism/Extensionality.agda
```diff
@@ -0,0 +1,23 @@
...
+
+open import Cat.Base
+
+import Cat.Morphism
+
+module Cat.Morphism.Extensionality {o ℓ} {C : Precategory o ℓ} where
+
+open Cat.Morphism C
+
+Extensional-≅
```
TOTBWF, line 11: Nice!!!!

src/Order/DCPO.lagda.md
```diff
@@ -337,15 +338,6 @@ module _ {o ℓ} {D E : DCPO o ℓ} where
     module E = DCPO E
   open is-directed-family
   open Total-hom
-
-  scott-path
```
TOTBWF, line 341: Very happy to see this go away :)

TOTBWF (approved):
This is amazing!!!

PR #276, opened 2023-10-28
chore: misc fixes

ncfavier:
In this week's edition of "chore: misc fixes", we

- fix a typo;
- use the [suggested](https://homotopytypetheory.org/book/) BibTeX entry for the HoTT book, which includes the URL and formats the "author" field properly (currently it looks like "Program 2013", first name "The Univalent Foundations");
- move the `overflow-x: auto` from https://github.com/plt-amy/1lab/pull/269 from `.Agda` to `pre.Agda` to exempt the sidebar module name from it, because *for some reason* that causes it to collapse vertically, making it invisible or barely visible on most pages.

### DIFF
--- a/src/1Lab/Path/IdentitySystem.lagda.md
+++ b/src/1Lab/Path/IdentitySystem.lagda.md
@@ -54,7 +54,7 @@ record
 open is-identity-system public
 ```
 
-As mentioned before, the data of an identity system gives is exactly
+As mentioned before, the data of an identity system gives us exactly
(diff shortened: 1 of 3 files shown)


plt-amy:
> because for some reason that causes it to collapse vertically, making it invisible or barely visible on most pages.

@ncfavier what browser are you using? i can't reproduce on firefox or chromium

ncfavier:
Reproduced on both firefox and chromium on both computers I have... don't you love CSS

PR #287, opened 2023-11-19
Classical Propositional Logic

plt-amy:
Reopen #286:

> # Description
>
> This PR defines classical propositional logic, proves completeness and soundness, defines conjunctive normal forms, and also includes a simple DPLL SAT solver. It also includes a bunch of misc. additions to modules involving `Fin`, `List`, and `Bool`.

### DIFF
--- a/src/Logic/Propositional/Classical.lagda.md
+++ b/src/Logic/Propositional/Classical.lagda.md
@@ -1,10 +1,11 @@
 <!--
 ```agda
 open import 1Lab.Prelude hiding (_∈_)
-open import Data.Fin using (Fin; fzero; fsuc; weaken; inject)
-open import Data.Nat
-open import Data.List hiding (_++_)
+
 open import Data.Bool
+open import Data.Fin using (Fin; fzero; fsuc; weaken; inject)
+open import Data.List hiding (_++_)
+open import Data.Nat
 open import Data.Sum
 
 open import Meta.Brackets
@@ -337,4 +338,16 @@ sound (¬-elim {P = P} p q) ρ hyps-true =
   absurd $ not-no-fixed (sound q ρ hyps-true ∙ sym (sound p ρ hyps-true))
 sound (dneg-elim {P = P} p) ρ hyps-true =
   sym (not-involutive (⟦ P ⟧ ρ)) ∙ sound p ρ hyps-true
+```
+
+## Satisfiability
+
+```agda
+is-satisfiable : Proposition Γ → Type _
+is-satisfiable {Γ = Γ} P = Σ[ ρ ∈ (Fin Γ → Bool) ] (⟦ P ⟧ ρ ≡ true)
+
+equisat : Proposition Γ → Proposition Δ → Type _
+equisat P Q =
+  (is-satisfiable P → is-satisfiable Q) ×
+  (is-satisfiable Q → is-satisfiable P)
 ```
--- a/src/Logic/Propositional/Classical.lagda.md
+++ b/src/Logic/Propositional/Classical.lagda.md
@@ -27,12 +27,10 @@ the following indexed inductive type.
 
 ```agda
 data Proposition (Γ : Nat) : Type where
-  atom : Fin Γ → Proposition Γ
-  “⊤” : Proposition Γ
-  “⊥” : Proposition Γ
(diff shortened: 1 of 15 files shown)


src/Logic/Propositional/Classical/CNF.lagda.md
```diff
@@ -0,0 +1,322 @@
...
+avoid-lit i (lit x) p = lit (avoid i x p)
+avoid-lit i (neg x) p = neg (avoid i x p)
+
+```
+-->
+
+
+## Semantics
+
+Like their non-normal form bretheren, expressions in CNF have a natural
```
ncfavier, line 100:
```suggestion
Like their non-normal form brethren, expressions in CNF have a natural
```

src/Logic/Propositional/Classical/SAT.lagda.md
```diff
@@ -0,0 +1,297 @@
...
+
+# Unit Propagation
+
+The algorithm we will use is a simplified version of the classic DPLL
+algorithm, which combines backtracking search with a mechanism for
+pruning the search space, known as **unit propagation**. The idea
+behind this is as follows: if we see an clause that contains a single
+literal $P$ in our expression, then we know that $P$ must be true.
+Furthermore, any clause containing $P$ can be deleted, as we know it must
+be true! Even better, we can remove any occurance of $\neg P$ from
```
ncfavier, line 55:
```suggestion
be true! Even better, we can remove any occurrence of $\neg P$ from
```

src/Logic/Propositional/Classical/SAT.lagda.md
```diff
@@ -0,0 +1,297 @@
...
+...   | no ¬sat-false = no λ (ρ , ρ-sat) →
+        Bool-elim (λ b → ρ fzero ≡ b → ⊥)
+          (λ ρ₀-true → unit-propagate-unsat (lit fzero) ϕs ¬sat-true (ρ , ρ-sat , ρ₀-true))
+          (λ ρ₀-false → unit-propagate-unsat (neg fzero) ϕs ¬sat-false (ρ , ρ-sat , ap not ρ₀-false))
+          (ρ fzero) refl
+```
+
+And that's it! Note that "classic" DPLL also includes a second rule
+known as "pure literal elimination". The idea here is that if a literal
+only shows up as negated or not negated, then we can delete all occurances
```
ncfavier, line 294:
```suggestion
only shows up as negated or not negated, then we can delete all occurrences
```

src/Logic/Propositional/Classical.lagda.md
```diff
@@ -0,0 +1,778 @@
...
+shift-prop-“¬” (weak σ) P = shift-prop-“¬” σ (bump-prop P)
+```
+</details>
+
+## Context inclusions
+
+We also want to be able to modify our collection of hypotheses.
+We do this by defining a notion of context inclusion, where
+an inclusion $\Gamma \to \Delta$ denotes that we can obtain every
+hypotheses in $\Delta$ dropping some hypotheses in $\Gamma$. We call
```
ncfavier, line 276:
```suggestion
hypotheses in $\Delta$ by dropping some hypotheses in $\Gamma$. We call
```

TOTBWF (dismissed):

src/Logic/Propositional/Classical.lagda.md
```diff
@@ -0,0 +1,778 @@
...
+sound (¬-elim {P = P} p q) ρ hyps-true =
+  absurd $ not-no-fixed (sound q ρ hyps-true ∙ sym (sound p ρ hyps-true))
+sound (dneg-elim {P = P} p) ρ hyps-true =
+  sym (not-involutive (⟦ P ⟧ ρ)) ∙ sound p ρ hyps-true
+```
+
+## Completeness
+
+As mentioned earlier, we often think of classical propositional logic as
+having a *single* semantics. This is largely justified by to the fact
```
ncfavier, line 571:
```suggestion
having a *single* semantics. This is largely justified by the fact
```

src/Logic/Propositional/Classical.lagda.md
plt-amy: That first one should be "hypothesis" too

ncfavier (approved):

PR #288, opened 2023-11-20
refactor: rename ∥-∥-proj to ∥-∥-proj!

TOTBWF:
# Description

We were often using `∥-∥-proj` with the tactic argument applied, so I renamed `∥-∥-proj` to `∥-∥-proj!`, and added a new function that takes an explicit proof of `is-prop`. This was requested by @favonia.


favonia:
1. Should we refactor `is-prop→equiv∥-∥`?
2. Should `∥-∥-proj!` be defined in terms of `∥-∥-proj`?

TOTBWF:
Good point; done!

favonia:
Let's hope the CI god will approve our prayer.

plt-amy (approved):
Very sensible, and you anticipated by question of "did you go looking for all the `{ap = ...` 🙂

PR #291, opened 2023-11-26
A macro for generating induction principles

ncfavier:
# Description

A macro for generating induction principles for (higher, indexed) inductive types.

We support inductive types in the form presented in "Pattern Matching Without K" by Cockx et al, section 3.1, with the addition of higher inductive types.

When generating an eliminator into (n-2)-types, we automatically omit the methods that follow from h-level (i.e. those corresponding to constructors with at least n nested `PathP`s).

This still needs polishing; if you find bugs, please report them! Here's one:

Defining *recursion* principles tends to create unsolved metas from the method arguments to the recursive calls that start with implicit Π types, because Agda instantiates them with metavariables. This could be solved by doing `λ {x y z} → mᵢ {x} {y} {z}`, but is there a better way?

Also needs optimisation; ~in particular, many `raiseTC` calls on things that are known to be unapplied variables could be avoided~. I am not very good at determining what hurts performance the most though, so any help is appreciated.

## Checklist

- [X] The imports of new modules have been sorted with `support/sort-imports.hs`.

### DIFF
--- a/src/1Lab/Reflection/Induction.agda
+++ b/src/1Lab/Reflection/Induction.agda
@@ -1,0 +1,290 @@
+open import 1Lab.Reflection.Subst
+open import 1Lab.HLevel.Retracts
+open import 1Lab.Reflection
+open import 1Lab.Type.Sigma
+open import 1Lab.Type.Pi
+open import 1Lab.HLevel
+open import 1Lab.Path
+open import 1Lab.Type
+
+open import Data.String.Show
+open import Data.Nat.Base
+
+open import Meta.Append
+
+module 1Lab.Reflection.Induction where
+
+open import Data.Maybe.Base public
+
+{-
+A macro for generating induction principles for (higher, indexed) inductive types.
+
+We support inductive types in the form presented in "Pattern Matching Without K"
+by Cockx et al, section 3.1, with the addition of higher inductive types.
+Reusing the terminology from that paper, the idea is to translate the constructors
+of a data type into corresponding *methods* along with spines that tell us how to
+apply them to build the eliminator's clauses.
+
+When generating an eliminator into (n-2)-types, we automatically omit the methods that
+follow from h-level (i.e. those corresponding to constructors with at least n
+nested `PathP`s).
+
+See 1Lab.Reflection.Induction.Examples for examples.
+-}
+
+pi-view-path : Term → Telescope
+pi-view-path (pi x (abs n y)) = (n , x) ∷ pi-view-path y
+pi-view-path (def (quote PathP) (_ h∷ lam _ (abs n y) v∷ _ v∷ _ v∷ [])) =
+  (n , argN (quoteTerm I)) ∷ pi-view-path y
+pi-view-path _ = []
+
+Replacements = List (Term × Term)
+
+subst-replacements : Subst → Replacements → Replacements
+subst-replacements s = map (×-map (applyS s) (applyS s))
+
+instance
+  Has-subst-Replacements : Has-subst Replacements
+  Has-subst-Replacements = record { applyS = subst-replacements }
+
+private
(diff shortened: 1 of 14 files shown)


src/1Lab/Reflection/Induction.agda
```diff
@@ -0,0 +1,290 @@
...
+
+  lookupR : Term → Replacements → Maybe Term
+  lookupR t [] = nothing
+  lookupR t@(con c _) ((con c' _ , r) ∷ rs) with c ≡? c'
+  ... | yes _ = just r
+  ... | no  _ = lookupR t rs
+  lookupR t@(var i _) ((var i' _ , r) ∷ rs) with i ≡? i'
+  ... | yes _ = just r
+  ... | no  _ = lookupR t rs
+  lookupR t (_ ∷ rs) = lookupR t rs
```
ncfavier, line 73: I tried to use `lookup` instead of this but the `Discrete-Term` instance seems to take ages (?)

We could also make `Replacements` a `List ((Name ⊎ Nat) × Term)` but then we'd only have `raise`, not `subst`

src/1Lab/Reflection/Induction.agda
```diff
@@ -0,0 +1,290 @@
...
+      module I = Induction D pars ixTel cs ind? hlevel? hide-method-args? hide-indices?
+  methodTel , αs , rs ← in-context (reverse (parTelH ++ motiveTel ++ H-LevelTel)) (I.methods ps P)
+  let baseTel = parTelH ++ motiveTel ++ hlevelTel ++ methodTel
+      DTel = raise (length hlevelTel + length methodTel) DTel
+      P = raise (length methodTel + length DTel) P
+      Pd = apply-tm* P (tel→args 0 (PTel DTel))
+      -- ∀ (ps : Γ) {ℓ} {P} (h : ∀ d → is-hlevel (P d) n)? (ms : methodTel) (d : DTel) → P d
+      elimT = unpi-view (baseTel ++ DTel) Pd
+  elimT' ← just <$> get-type elim <|> nothing <$ declare (argN elim) elimT
+  for elimT' (unify elimT) -- Give a nicer error if the types don't match
```
ncfavier, line 258: "amusingly", using `traverse` here instead of `for` makes Agda hang

src/1Lab/Reflection/Induction.agda
plt-amy: I think that makes sense, `Discrete-Term` is comparing quite a few more things than your `lookup`..

ncfavier: I mean I tried it with `con₀` and `var₀`, so you'd think it would only compare the `Name`/`Nat`s and instantly see whether the lists are equal, but instead it seemed to hang forever.

src/1Lab/Reflection/Induction.agda
plt-amy: It doesn't hang for me, but it does block a billion metas from being solved. I don't want to think about it

src/1Lab/Reflection/Induction.agda
```diff
@@ -0,0 +1,293 @@
...
+        (just m) → do
+          -- The type of the method is solvable by hlevel (i.e. contractible):
+          -- we can omit that type from the telescope and replace the method with
+          -- a call to `hlevel 0 .centre`.
+          let rs = (c' , m) ∷ rs
+          go ps P rs cs <&> ×-map₁₂ id (α ∷_)
+        nothing → do
+          -- Otherwise, add m : T to the telescope and replace the corresponding
+          -- constructor with m henceforth.
+          let method = "m" <> show i
```
plt-amy, line 207:
```suggestion
          cname ← render-name c
          let method = "m" <> cname
```

PR #294, opened 2023-11-29
chore: bump Agda

plt-amy:
The changes to instance ordering did actually reveal some less-than-defined behaviour in the hlevel tactic: if the goal type was blocked on a meta, it might commit to a solution too early, blocking another tactic (or instance search) from making progress. 

This came to head in the `Abelianisation` module, where `hlevel`, `extensionality`, and a `Funlike` instance all had to cooperate to solve the funlike instance.

### DIFF
--- a/src/1Lab/Reflection/HLevel.agda
+++ b/src/1Lab/Reflection/HLevel.agda
@@ -196,6 +196,27 @@ private
     lv ← wait-just-a-bit lv
     pure (ty , lv)
 
+  -- Wait for the "principal argument" of a term, i.e. the (principal
+  -- argument of) the first visible argument in the spine of a
+  -- definition.
+  wait-principal-arg : Term → TC ⊤
+  wait-principal-arg topl = go topl where
+    go : Term → TC ⊤
+    go* : List (Arg Term) → TC ⊤
+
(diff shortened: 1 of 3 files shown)


SquidDev (approved):
👍 from me - had a couple of comments/questions, but happy to be merged as-is:

Was the `-f debug` added for debugging the tactic issues, or is this something we want to keep permanently? I know it has some performance cost, so just wanted to check.

I also wonder if it's worth removing the `noProfile` - I know I sometimes find it useful to recompile 1lab with profiling enabled.

plt-amy:
> Was the `-f debug` added for debugging the tactic issues, or is this something we want to keep permanently? I know it has some performance cost, so just wanted to check.

Both, I think :) Tactics do print their debug output even when Agda doesn't, of course; but debugging tactics pretty often ends up in me fiddling with the elaborator's verbosity, too. We could get rid of the few verbosity flags scattered around, though!

> I also wonder if it's worth removing the `noProfile` - I know I sometimes find it useful to recompile 1lab with profiling enabled.

Good call, I'll kick off a build.

SquidDev (approved):
Thank you!

PR #300, opened 2023-12-02
Search revamp

plt-amy:
Just experimenting with some improvements to the search box


support/web/js/theme.tsx
```diff
@@ -0,0 +1,132 @@
+import { JSX, type Content } from "./lib/jsx";
+import { type Theme, themeSetting, equationSetting, Setting, hiddenCodeSetting, footnoteSetting, serifFontSetting } from "./lib/settings";
+
+function Button(props: { label: Content, icon?: string, click: ((ev: MouseEvent) => void) | string, ['class']?: string }): HTMLButtonElement {
+  const el: HTMLElement =
+    <button class={`labelled-button ${props['class']}`}>
+      <span class="icon" style={`background-image: url('/static/icons/${props.icon}.svg');`}></span>
```
SquidDev, line 7: This won't render correctly on the GH preview, as it uses an absolute path rather than a relative one. Could we prepend `window.baseURL` to it?

support/web/js/theme.tsx
```diff
@@ -0,0 +1,132 @@
...
+      <span class="hover-label">{props.label}</span>
+    </button>;
+
+  el.addEventListener("click", (ev) => {
+    if (typeof props.click === 'string') {
+      window.location.href = props.click;
+      return;
+    }
```
SquidDev, line 15: I think we should probably use a `a` tag when the click accessor is a string (it should be possible to style the two tags the same). This breaks a load of functionality of links (e.g. open in new tab).

SquidDev (changes requested):
Oh, this is really nice! I had a couple of nitpicks about the buttons, but otherwise this looks good to me!

ncfavier:
Where'd the options go? (Serif, hidden code, etc.)

The search results are a bit small, and the module names on the right of the search results are definitely too small.

plt-amy:
> Where'd the options go? (Serif, hidden code, etc.)

support/web/js/theme.tsx
```diff
@@ -1,20 +1,28 @@
 import { JSX, type Content } from "./lib/jsx";
 import { type Theme, themeSetting, equationSetting, Setting, hiddenCodeSetting, footnoteSetting, serifFontSetting } from "./lib/settings";
 
+// This is pretty evil, but a loose <script> tag assigns these to the
+// window object in the HTML template.
+const links: { baseURL: string, source: string } = window as any;
```
SquidDev, line 6: We could add a `.d.ts` file which defines these, but probably not worth it.

support/web/js/theme.tsx
```diff
@@ -1,20 +1,28 @@
...
+
 function Button(props: { label: Content, icon?: string, click: ((ev: MouseEvent) => void) | string, ['class']?: string }): HTMLButtonElement {
+  const style = `background-image: url('${links.baseURL}/static/icons/${props.icon}.svg');`;
+
+  const ic: HTMLElement = typeof props.click === 'string' ?
+    <a class="icon" style={style} href={props.click} /> :
+    <span class="icon" style={style} />;
```
SquidDev, line 13: Oh, I see what you meant on IRC. Oh, this is really sad, but I don't have a better solution.

SquidDev (approved):
:heart:  Thank you!

PR #314, opened 2023-12-14
`Sets` is not self-dual

mmcqd:
# Description

This PR contains a proof that `Sets` is not self-dual, which is placed in the new `Cat/Instances/Sets/Counterxamples/` folder. To use in this proof, it also defines strict initial objects, proves `Sets ^op` is a category, and fixes a typo.

This is my first contribution so I expect my prose/code may need some improvement, I'm open to any criticism!

## Checklist

- [x] The imports of new modules have been sorted with `support/sort-imports.hs`.


src/Cat/Instances/Sets/Counterexamples/SelfDual.lagda.md
```diff
@@ -0,0 +1,120 @@
...
+open import Data.Bool
+```
+-->
+
+```agda
+module Cat.Instances.Sets.Counterexamples.SelfDual {ℓ} where
+```
+# Sets is not self-dual
+
+We show that the category of Sets is not self-dual, that is, there cannot exit a path between `Sets`{.Agda} and `Sets`{.Agda} `^op`{.Agda}.
```
ncfavier, line 17:
```suggestion
We show that the category of sets is not self-dual, that is, there cannot exist a path between `Sets`{.Agda} and `Sets`{.Agda} `^op`{.Agda}.
```

Also I think $\mathbf{Sets}$ and $\mathbf{Sets}^\mathrm{op}$ should be typeset in math throughout, i.e `$\Sets$` and `$\Sets\op$`.

src/Cat/Instances/Sets/Counterexamples/SelfDual.lagda.md
```diff
@@ -0,0 +1,120 @@
...
+Sets^op-initial .Initial.bot = el! (Lift _ ⊤)
+Sets^op-initial .Initial.has⊥ x = hlevel!
+```
+<!--
+```agda
+_ = ⊥
+``` 
+-->
+
+Now we can observe an interesting property of the initial object of `Sets`{.Agda}: every morphism into it is in fact an *iso*morphism.
```
ncfavier, line 49: Perhaps worth mentioning that this property is called having a [strict initial object](https://ncatlab.org/nlab/show/strict+initial+object).

This is an instance of the wider phenomenon of [van Kampen colimits](https://ncatlab.org/nlab/show/van+Kampen+colimit), if you're interested. (I am planning to add those to 1lab soon™.)

src/Cat/Instances/Sets/Counterexamples/SelfDual.lagda.md
plt-amy: :+1: on using `\Sets`

src/Cat/Instances/Sets/Counterexamples/SelfDual.lagda.md
mmcqd: Oh that's cool, yes I can mention this

mmcqd: done

src/Cat/Diagram/Initial.lagda.md
```diff
@@ -79,13 +80,41 @@ Additionally, if $C$ is a category, then the space of initial objects is
...
+
+[strict]: https://ncatlab.org/nlab/show/strict+initial+object
+[van Kempen colimits]: https://ncatlab.org/nlab/show/van+Kempen+colimit
+
+
+```agda
+is-strict-initial : Initial → Type _
+is-strict-initial i = ∀ X → Hom X (i .bot) → X ≅ i .bot
+
+record StrictInitial : Type (o ⊔ h) where
```
plt-amy, line 109:
```suggestion
record Strict-initial : Type (o ⊔ h) where
```

src/Cat/Diagram/Initial.lagda.md
```diff
@@ -79,13 +80,41 @@ Additionally, if $C$ is a category, then the space of initial objects is
...
+  field
+    initial : Initial
+    has-is-strict : is-strict-initial initial
+```
+
+Strictness is a property of initial objects
```
plt-amy, line 115:
```suggestion
Strictness is a property of, not structure on, an initial object.
```

src/Cat/Instances/Sets.lagda.md
```diff
@@ -89,3 +91,41 @@ We then use [univalence for $n$-types] to directly establish that $(A
...
     (A Sets.≅ B)    ≃∎
 ```
+
+## Sets^op is also a category
+
+```agda
+  import Cat.Reasoning (Sets ℓ ^op) as Sets^op
+```
+First we show that isomorphism is invariant under `^op`{.Agda}.
+```agda
```
plt-amy, line 101: To be safe it's best to always leave a blank line between code blocks and text

src/Cat/Diagram/Initial.lagda.md
```diff
@@ -79,13 +79,41 @@ Additionally, if $C$ is a category, then the space of initial objects is
...
       {A = Hom (Univalent.iso→path ccat (⊥-unique x1 x2) i) _})
     (x1 .has⊥ ob) (x2 .has⊥ ob) i
 ```
+
+## Strictness
+
+An initial object is said to be *[strict]* if every morphism into it is an *iso*morphism.
+This is a categorical generalization of the fact that if one can write a function $X \to \bot$ then $X$ must itself be empty.
+
+This is an instance of the more general notion of [van Kempen colimits].
```
ncfavier, line 98: van K**a**mpen.

plt-amy (dismissed):

src/Cat/Diagram/Initial.lagda.md
```diff
@@ -79,13 +79,41 @@ Additionally, if $C$ is a category, then the space of initial objects is
...
+
+This is an instance of the more general notion of [van Kampen colimits].
+
+[strict]: https://ncatlab.org/nlab/show/strict+initial+object
+[van Kampen colimits]: https://ncatlab.org/nlab/show/van+Kampen+colimit
+
+
+```agda
+is-strict-initial : Initial → Type _
+is-strict-initial i = ∀ x → Hom x (i .bot) → x ≅ i .bot
```
ncfavier, line 106: Nitpick: I would prefer to use the usual definition, "every morphism x → ⊥ is invertible" (`(f : Hom x (i .bot)) → is-invertible f`), because it is straightforward to get your definition from this one, whereas I am not immediately sure about the converse (I haven't thought about it very hard).

ncfavier: Well I guess if x ≅ ⊥ then morphisms out of x are unique, so f is also invertible; still I would rather not have to argue that anywhere.

mmcqd: I just learned about these today so I have no strong feelings, the nlab uses the isomorphism definition, is the `is-invertible` definition the one that falls out of van kampen colimits? As you say I think these end up being the same.

ncfavier: Invertible is another name for isomorphism in this context, what differs is whether you get a *different* function than what you put in, underlying the isomorphism. The definition in nlab gives you the same f, as does the one I suggest.

(The diagram there is misleading, but note that f appears on both sides.)

mmcqd: Ahhh, I see. `is-invertible` is the same as `is-iso` yeah? I think you're right that your suggestion is better.

ncfavier: `is-iso` is taken by isomorphisms in the sense of HoTT (i.e. in the "ambient" category of types and functions). `is-invertible` is the general categorical notion.

mmcqd: and this makes `is-strict-initial-is-prop` solvable by `hlevel!`, nice. Thanks for the suggestion

ncfavier (approved):

plt-amy:
Thank you!

PR #315, opened 2023-12-14
chore: only build files tracked by git

ncfavier:


plt-amy:
i like having the wip files built though

plt-amy (approved):

PR #324, opened 2023-12-19
defn: two point circle

KevOrr:
# Description

A new module for an HIT which is essentially S^1 but with two point constructors, along with a proof of equivalence with S^1 and an example of stealing theorems from S^1 given this equivalence. Funnily enough, while cleaning it up, I realized it was actually much simpler than what I shared in Discord, and the only lemmas it needed about composition ended up already existing in other modules.

## Checklist

- [X] The imports of new modules have been sorted with `support/sort-imports.hs`.

### DIFF
--- a/src/Homotopy/Space/Circle/TwoPoint.lagda.md
+++ b/src/Homotopy/Space/Circle/TwoPoint.lagda.md
@@ -1,0 +1,163 @@
+<!--
+```agda
+open import 1Lab.Prelude
+
+open import Data.Int.Universal
+open import Data.Int
+
+open import Homotopy.Space.Circle hiding (ΩS¹≃integers)
+```
+-->
+
+```agda
+module Homotopy.Space.Circle.TwoPoint where
+```
+
+# Spaces: The circle with two point constructors
+
+We can add additional points onto the [[circle]], and still end up with
+a space that is homotopy-equivalent to `S¹`{.Agda}.
+
+~~~{.quiver}
+\begin{tikzpicture}
+\node[draw,circle,label=below:{$\rm{south}$},fill,outer sep=0.1cm, inner sep=0pt, minimum size=0.1cm] (south) at (0, -1) {};
+\node[draw,circle,label=above:{$\rm{north}$},fill,outer sep=0.1cm, inner sep=0pt, minimum size=0.1cm] (north) at (0, 1) {};
+\draw[->, bend right=85, distance=1.2cm] (south) to node[right] {$\rm{east}\ i$} (north);
+\draw[->, bend right=85, distance=1.2cm] (north) to node[left] {$\rm{west}\ i$} (south);
+\end{tikzpicture}
+~~~
+
+```agda
+data S¹₂ : Type where
+  south : S¹₂
+  north : S¹₂
+  east : south ≡ north
+  west : north ≡ south
+```
+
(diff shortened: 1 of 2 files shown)


src/Homotopy/Space/Circle/TwoPoint.lagda.md
```diff
@@ -0,0 +1,163 @@
...
+S¹₂→S¹ south = base
+```
+
+But what shall we map `north`{.Agda} (and by extension, `east`{.Agda}
+and `west`{.Agda}) to? If the interval type were literally the unit
+interval of the real numbers, $[0,1]$, then we could map `north`{.Agda}
+to the point halfway along `loop`{.Agda}. However, the only two points
+along the interval that we can name are `i0`{.Agda} and `i1`{.Agda}.
+Thus, we need to map `north`{.Agda} to either `base`{.Agda}, `loop i0`,
+or `loop i1`, since these are the only terms of type `S¹`{.Agda} that we
```
KevOrr, line 75: > since these are the only terms [...] we can name

Not technically true, cause we have closed `hcomp` terms in S^1 that don't reduce (IIUC). Not sure how rigorous this should be though

ncfavier:
This seems redundant with our existing [proof](https://1lab.dev/Homotopy.Space.Sphere.html#1909) that the suspension of 2 points is equivalent to S¹?

src/Homotopy/Space/Circle/TwoPoint.lagda.md
ncfavier: Good enough for intuition, but note that `loop i0` and `loop i1` are `base`, so there's really only one normal term (ignoring empty hcomps).

plt-amy:
Hum, the maths is redundant, but the explanation are better. Can we replace that for this?

Lavenza:
[Pull request preview](https://preview.1lab.dev/324)

src/Homotopy/Space/Circle/TwoPoint.lagda.md
```diff
@@ -0,0 +1,163 @@
...
+can name. However, all three of these terms are definitionally
+`base`{.Agda}! So let's use that.
+
+```agda
+S¹₂→S¹ north = base
+```
+
+The question now arrises how we map `east`{.Agda} and `west`{.Agda} onto
```
ncfavier, line 83:
```suggestion
The question now arises how we map `east`{.Agda} and `west`{.Agda} onto
```

src/Homotopy/Space/Circle/TwoPoint.lagda.md
KevOrr: It does say that those are all definitionally `base`, but yeah I wasn't sure how much people think up irreducible hcomps when thinking about members of HITs

KevOrr:
> This seems redundant with our existing [proof](https://1lab.dev/Homotopy.Space.Sphere.html#1909) that the suspension of 2 points is equivalent to S¹

Oh yeah, of _course_ that proof would have to have a similar approach. Yeah, this shouldn't be a new module then since this HIT is obviously equivalent to `Sⁿ⁻¹ 2`

> Can we replace that for this?

Yes, do you want me to transport (heh) these explanations to that proof in `Homotopy.Space.Sphere.html`?

PR #327, opened 2023-12-25
More instances of orders

plt-amy:
Includes poset facts for `Nat`/`Int` and @mmcqd's work on poset constructions. Also includes definition of total orders + proof that these have meets and joins.

"More instances of orders", they say, in the middle of rewriting everything to do with integers... The initial ring is still using the higher-inductive ℤ, but everything else is ported (there wasn't much).


Lavenza:
[Pull request preview](https://preview.1lab.dev/327)

favonia:
Is `+ℤ-injectiver` used anywhere? It's needed in our own development, but it seems a bit strange in the generic library without `+ℤ-injectivel`.

plt-amy:
I added it only because it was part of the `Mugen.Data.Int` module I was upstreaming; there's not much about integers in the 1Lab yet, so it's not used. I guess I could add `+ℤ-injectivel` as well

Lavenza (approved):

favonia:
Speaking of upstreaming, let me note here for the record that the credit of the original file belongs to @TOTBWF and me, BUT @plt-amy has pretty much rewritten the whole thing from scratch.

Issue #147, opened 2022-10-22
On higher zoom levels (125%+), the sidebar disappears with no way to access it


Dessertion:
The sidebar disappearing is intentional as a feature, which makes sense since it prevents the main content from getting too squished. However, a side-effect of the current implementation is that there is no way to directly access the sidebar while at higher zooms and more importantly **no indication** that there even is a sidebar (particularly problematic when a user has default page zoom set above 125%).

A solution to both issues here would be perhaps to introduce some sort of dropdown/popup/something (sorry, not a UI designer). Perhaps a "hamburger menu"-esque thing on the top bar? Or maybe a pullout drawer kind of feature. 

Thanks!

solson:
+1 for this. I don't even use zoom, but I tend to lay out my windows taking up halves of the screen, so my browser window tends to be portrait rather than landscape aspect ratio, and the sidebar is again hidden with no way to access it or indication that it exists.

marshall-lee:
As for me, the problem is the opposite. I haven't known that the website has the sidebar until I _zoomed out_.
MacBook Pro 2015, reproduces both in Firefox and Chromium.

PR #338, opened 2024-01-08
web: Tufte-style sidenotes and justified text

plt-amy:
Not much to say here, just gotta wait for the preview to render, I suppose.


Lavenza:
[Pull request preview](https://preview.1lab.dev/338)

ncfavier:
The sidebar is gone for me in the preview :(

SquidDev:
Is that another instance of #147, just now more visible as the breakpoints have changed again?

support/web/js/lib/settings.tsx
```diff
@@ -51,9 +52,12 @@ export type Theme = 'light' | 'dark' | 'system';
 export const equationSetting   = new Setting<boolean>("equations",   false)
   .onChange((v) => v ? clz.add("show-equations") : clz.remove("show-equations"));
 
-export const serifFontSetting  = new Setting<boolean>("sans_serif",  false)
+export const serifFontSetting  = new Setting<boolean>("sans_serif", false)
   .onChange((v) => v ? clz.remove("sans-serif") : clz.add("sans-serif"));
```
SquidDev, line 56: Wait, is this option the wrong way round? When `sans_serif` is false, we add the `sans-serif` font!

I'm not sure what the intended default is, but we should possibly update the CSS to match it - currently that prefers serif fonts over sans serif ones when JS is disabled.

SquidDev (approved):
This is really nice! It's depressing we need to rely on JS at all here, I'm amazed how well it works!

The only problem I've noticed is that footnotes aren't visible when JS is disabled. We could fix it by adding a CSS class to the body from Javascript to enable the footnotes, but I'm not sure either of us can face wrangling the CSS for that, so inclined to say ship it as is :D.

support/web/css/vars.scss
```diff
@@ -105,3 +105,6 @@ $light-button-hover-bg: $zinc-200;
 
 $dark-text-bg: #23272E;
 $dark-text-fg: #ABB2BF;
+
+// "Desktop layout": navigation bar, content, sidenotes.
+$desktop-layout-width: 110rem;
```
ncfavier, line 110: Given https://github.com/plt-amy/1lab/issues/147, we should keep this low. The minimum for my 1920x1080, 120 DPI laptop screen seems to be 1920*96/120/16 = 96rem, minus one for safety:

```suggestion
$desktop-layout-width: 95rem;
```

ncfavier:
The targets of footnote links (like `#fn1`) are now missing, so clicking them doesn't do anything. Is that intended?

plt-amy:
@ncfavier Yeah, it's intentional. Since browsers won't move you to a hash that's not painted, clicking the footnote references on the desktop layout does nothing. But when the screen is narrow and the footnotes render again, the footnote references just work, without us having to shuffle the DOM, or override `click` for the anchors, etc.

Lavenza (approved):

support/web/css/code.scss
```diff
@@ -13,110 +12,115 @@
...
 }
 
-code, pre, .sourceCode {
-  font-family: 'Julia Mono';
-  font-size: var(--code-font-size);
-  font-weight: 500;
-}
+code, pre, .sourceCode { @include monospaced; }
+
+code, pre { width: min-content; }
```
ncfavier, line 17: Uuuh

screenshot-2024-01-27-094743

plt-amy: open a new issue

PR #340, opened 2024-01-11
Small things

ncfavier:
# Description

See commits

## Checklist

### DIFF
--- a/src/Cat/Allegory/Reasoning.lagda.md
+++ b/src/Cat/Allegory/Reasoning.lagda.md
@@ -15,7 +15,7 @@ module Cat.Allegory.Reasoning {o ℓ ℓ'} (A : Allegory o ℓ ℓ') where
 ```agda
 open Allegory A public
 open Cat.Reasoning (A .Allegory.cat)
-  hiding (module HLevel-instance ; Ob ; Hom ; Hom-set ; id ; idl ; idr ; assoc ; _∘_ ; Mor ; Hom→Mor ; Mor-path)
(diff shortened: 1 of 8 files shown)


Lavenza:
[Pull request preview](https://preview.1lab.dev/340)

src/Cat/Displayed/Instances/Slice.lagda.md
```diff
@@ -228,6 +230,33 @@ pullback→cartesian {x} {y} {x'} {y'} {f} {f'} pb = cart where
     pb.unique (sym (m' .commute)) (ap to x)
 ```
 
+The same is true for *weakly* cartesian morphisms, by an equally
+straightforward computation:
+
+```agda
```
TOTBWF, line 236: Doesn't this follow from https://1lab.dev/Cat.Displayed.Cartesian.Weak.html#4549?

ncfavier: It does if you have all pullbacks. I needed this to show that cartesian sections of pullbacks of the codomain fibration are equivalent to equifibered transformations, and there it seems unnecessary to ask for all pullbacks. (And the reason I need the *weakly* cartesian version is that change of base only preserves weakly cartesian morphisms in the direction i'm interested in.)

I concede that this is not very interesting though, so maybe it should be hidden.

TOTBWF: That makes sense! I think these proofs are still interesting, though I'd maybe add some prose explaining why these proofs exist so other people don't ask the same question!

plt-amy (approved):

PR #344, opened 2024-01-17
Day convolution

plt-amy:
20 coherences per second. Does not actually prove the universal property of `PSh C, _⊗ᴰ_` yet — have mercy.

### DIFF
--- a/src/Cat/Monoidal/Instances/Day.lagda.md
+++ b/src/Cat/Monoidal/Instances/Day.lagda.md
@@ -1,0 +1,357 @@
+<!--
+```agda
+{-# OPTIONS -vtc.decl:5 -vtactic.extensionality:10 #-}
+open import 1Lab.Rewrite
+
+open import Cat.Functor.Naturality
+open import Cat.Functor.Hom
+open import Cat.Diagram.Coend.Sets
+open import Cat.Instances.Product
+open import Cat.Diagram.Coend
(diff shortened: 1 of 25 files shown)


Lavenza:
[Pull request preview](https://preview.1lab.dev/344)

src/Cat/Monoidal/Instances/Day.lagda.md
```diff
@@ -0,0 +1,630 @@
...
+```
+
+<!--
+```agda
+    factor-day
+      : ∀ {i a b} {W : Cowedge (Day-diagram i)} {h : Hom i (a ⊗ b)} {x : X ʻ a} {y : Y ʻ b}
+      → factor W (day h x y) ≡rw W .ψ (a , b) (h , x , y)
+    factor-day = idrw
+
+  {-# REWRITE factor-day #-}
```
TOTBWF, line 291: I'm assuming this is needed as `day` is `opaque`? Might be good to add a note here explaining this.

TOTBWF (approved):
Looks good (though very painful!). Only remarks are that we should document the `REWRITE` in `Cat.Monoidal.Instance.Day`, and that the index should be updated.

I've got some more motivation I can add from the internal language POV (Day gives you context splitting representations of linear type theories), but I can do that in a seperate PR.

TOTBWF (approved):

PR #357, opened 2024-01-31
chore: bump to agda/agda#7077

plt-amy:
I wrote an Agda feature so we could have nice `_∈_` so here's my implementation of a nice `_∈_`. The only file that broke with the change to instance selection was `Cat.Instances.Discrete`, where a (non-confluent) rewrite rule no longer fired for one of the `F-id` proofs.

The diff is large mostly because I went around making some things prettier also. Remarkably *every single use-site* of both `ext` and `hlevel!` survived as is. We're pretty good at writing tactics that block properly :relieved:

### DIFF
--- a/src/1Lab/Underlying.agda
+++ b/src/1Lab/Underlying.agda
@@ -1,5 +1,6 @@
 open import 1Lab.HLevel.Retracts
 open import 1Lab.HLevel.Universe
+open import 1Lab.HIT.Truncation
 open import 1Lab.Resizing
 open import 1Lab.HLevel
 open import 1Lab.Path
@@ -64,16 +65,6 @@ from-is-true
   → ⌞ P ⌟
 from-is-true prf = subst ⌞_⌟ (sym prf) (hlevel 0 .centre)
 
--- Generalised "membership" notation.
-_∈_ : ∀ {ℓ ℓ'} {A : Type ℓ} {P : Type ℓ'} ⦃ u : Underlying P ⦄
-    → A → (A → P) → Type (u .ℓ-underlying)
-x ∈ P = ⌞ P x ⌟
-
--- Generalised "total space" notation.
-∫ₚ
-  : ∀ {ℓ ℓ'} {X : Type ℓ} {P : Type ℓ'} ⦃ u : Underlying P ⦄
-  → (X → P) → Type _
-∫ₚ P = Σ _ (_∈ P)
 
 -- Notation class for type families which are "function-like" (always
 -- nondependent). Slight generalisation of the homs of concrete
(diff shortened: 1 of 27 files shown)


Lavenza:
[Pull request preview](https://preview.1lab.dev/357)

src/Cat/Diagram/Sieve.lagda.md
```diff
@@ -76,7 +76,8 @@ action is given by composition, as with the $\hom$ functor.
 
 ```agda
 to-presheaf : Sieve → PSh.Ob
-to-presheaf sieve .F₀ d = el! (Σ[ f ∈ C.Hom d c ] (f ∈ sieve .arrows))
+to-presheaf sieve .F₀ d .∣_∣   = Σ[ f ∈ C.Hom d c ] (f ∈ sieve .arrows)
```
TOTBWF, line 79: How feasible is membership for sieves?

src/Cat/Functor/WideSubcategory.lagda.md
```diff
@@ -33,84 +34,103 @@ To start, we package up all the data required to define a wide
...
+          → P f → P g → P (f C.∘ g)
 ```
 
 Morphisms of wide subcategories are defined as morphisms in $\cC$ where
 $P$ holds.
 
+<!--
+```agda
+instance
+  Membership-Wide-subcat
```
TOTBWF, line 54: This is so nice...

src/1Lab/Underlying.agda
```diff
@@ -134,3 +125,48 @@ _ʻ_
   → ⦃ _ : Funlike F ⦄ {a : A} {b : B} ⦃ _ : Underlying ⌞ b ⌟ ⦄
   → F a b → ⌞ a ⌟ → Type _
 F ʻ x = ⌞ F # x ⌟
+
+record Membership {ℓ ℓ'} (A : Type ℓ) (ℙA : Type ℓ') ℓ'' : Type (ℓ ⊔ ℓ' ⊔ lsuc ℓ'') where
+  field
+    _∈_ : A → ℙA → Type ℓ''
```
TOTBWF, line 131: Mentioned this elsewhere, but would it make sense to bundle `A` in the record? That way instance resolution can only occur on the collection type, which feels like it might lead to less yellow

plt-amy: Pasting from Discord:

> I can't think of a reason to *allow* `x ∈ X` and `y ∈ X` when `x` and `y` have different types, but I don't see a reason to forbid it, either; And since Agda doesn't do typeclass generalisation, we're not going to get any improvement from changing the element type to be associated.

Also, Agda is quite happy to select an instance `Xm : Membership ?0 X _`, as soon as it's unambiguous; so there aren't any cases where instance selection _doesn't_ work with the class as-is, but would work with an associated type (having the associated type might even make it *harder*, since it'd cause the instance metas to get eta-expanded)

src/Algebra/Group/Subgroup.lagda.md
```diff
@@ -121,19 +126,25 @@ giving the element $xy$ in the fibre over $ab$.
...
+Extensional-Σ-∥∥ ⦃ sa ⦄ .Pathᵉ (x , _) (y , _) = sa .Pathᵉ x y
+Extensional-Σ-∥∥ ⦃ sa ⦄ .reflᵉ (x , _) = sa .reflᵉ x
+Extensional-Σ-∥∥ ⦃ sa ⦄ .idsᵉ .to-path p = Σ-prop-path! (sa .idsᵉ .to-path p)
+Extensional-Σ-∥∥ ⦃ sa ⦄ .idsᵉ .to-path-over p = sa .idsᵉ .to-path-over p
+
+instance
+  Extensionality-Σ-∥∥
+    : ∀ {ℓ ℓ'} {A : Type ℓ} {B : A → Type ℓ'} → Extensionality (Σ[ a ∈ A ] ∥ B a ∥)
+  Extensionality-Σ-∥∥ = record { lemma = quote Extensional-Σ-∥∥ }
+
```
plt-amy, line 141: Ope, I was experimenting with this instance but it should not live here.

src/Cat/Diagram/Sieve.lagda.md
plt-amy: The module would need some shuffling so that `C` and `c` can be invisible arguments to the instance, but it should be perfectly fine

ncfavier (approved):

plt-amy:
There's some uncertainty upstream about how exactly we'll go about getting instance search to support things like `Membership`, so I'll close this for now — we'll get this eventually, though, I promise :sweat_smile:

PR #361, opened 2024-02-17
Cleanup, rewrite 1Lab.Equiv, the opposite of a fibration

plt-amy:
The bulk of this PR is janitorial changes:

* All the instances of `F₀`/`Precategory.Ob`/`η` appearing as *prefix* projections are gone — at least those that could reasonably be identified with an almost-regular expression;

* Some path helpers e.g. `Element-hom-path` and `Nat-path` are entirely subsumed by the `ext` tactic, so I replaced those; and

* I went ahead with the `Membership` class, and just added the few type annotations required to get Agda to pick the right instances.

But pay attention to the following nontrivial work:

* The prose in `1Lab.Equiv` was basically entirely rewritten. I hope that the new text serves as a better introduction; to accompany I formalised the failure of "univalence-but-with-isos" — isovalence — so we have something concrete to point to as the answer for "why don't isos work?".

* The construction of the opposite fibration finally escaped the purgatory that is my wip directory :sweat_smile: I cleaned up the proofs substantially. The key insight was to work basically entirely in the fibre categories, only looking at `E` as a whole to establish a connection between transport in the domain and structural maps in the fibres.


Lavenza:
[Pull request preview](https://preview.1lab.dev/361)

src/Cat/Displayed/Instances/Opposite.lagda.md
```diff
@@ -0,0 +1,356 @@
...
+
+Since the theory of [[fibrations]] over $\cB$ behaves like ordinary
+category theory over $\Sets$, we expect to find $\cB$-indexed analogues
+of basic constructions such as functor categories, product categories,
+and, what concerns us here, *opposite* categories. Working at the level
+of [[displayed categories]], fix a fibration $\cE \liesover \cB$; we
+want to construct the fibration which is classified by
+
+$$
+\cB \xto{\cE^*(-)} \Cat \xto{(-)\op} \Cat\text{,}
```
ncfavier, line 65: Should one of these be $\mathfrak{Cat}^{\text{co}}$? I suppose since $\mathcal{B}$ is locally discrete its fibres might as well be valued in $\mathfrak{Cat}^{\text{co}}$.

src/Cat/Displayed/Instances/Opposite.lagda.md
```diff
@@ -0,0 +1,356 @@
...
+<!--
+```agda
+_^op' : Displayed B o' ℓ'
+_^op' .D.Ob[_] = Ob[_]
+_^op' .D.Hom[_]     f x y = Hom[ id ] (f ^* y) x
+_^op' .D.Hom[_]-set f x y = Hom[_]-set _ _ _
+_^op' .D.id'  = π
+_^op' .D._∘'_ = _∘,_
+```
+-->
```
ncfavier, line 246: This is the core of the definition, it should not be hidden!

src/Cat/Displayed/Instances/Opposite.lagda.md
```diff
@@ -0,0 +1,356 @@
...
+
+Since we defined the displayed morphism spaces directly in terms of
+$\cE$'s base change, it's not surprising that $\cE_{\rm{op}}$ is itself
+a [[fibration]]. Indeed, if we have $f : a \to b$ and $y \liesover b$, a
+Cartesian lift of $f^*_{\rm{op}}(y)$ *in the opposite* boils down to
+asking for something to fit into the question mark in the diagram
+
+~~~{.quiver}
+\[\begin{tikzcd}[ampersand replacement=\&]
+  {f^*(y)} \&\& {f^*y} \&\& y \\
```
ncfavier, line 327: Can we pick one of $f^\*y$, $f^\*(y)$ and $f^*_\text{op}(y)$?

plt-amy: well, one of those isn't like the others, but we can probably consistently drop the parentheses.

src/Cat/Displayed/Instances/Opposite.lagda.md
plt-amy: disagree — composition is defined way up at the start of the module, there's not really a reason to include putting it in a record

ncfavier: Can I bargain for `Ob[_]` and `Hom[_]`? I don't think the reader should have to infer what they are from the prose, even if it's obvious.

ncfavier (approved):

PR #372, opened 2024-04-02
equivalence implies identity system

ncfavier:
Remove the unnecessary hypothesis that the equivalence preserves refl.

## Checklist

### DIFF
--- a/src/1Lab/Path/IdentitySystem.lagda.md
+++ b/src/1Lab/Path/IdentitySystem.lagda.md
@@ -268,39 +268,16 @@ is-identity-system-is-prop {A = A} {R} {r} =
     from : (∀ x → is-contr (Σ A (R x))) → is-identity-system R r
     from x = contr→identity-system (x _)
 
-    sys : ∀ (l : ∀ x → is-contr (Σ A (R x))) a b (s : R a b) (i j : I)
-        → Partial (∂ i ∨ ~ j) (Σ A (R a))
-    sys l a b s i j (j = i0) = l a .centre
-    sys l a b s i j (i = i0) = l a .paths (a , r a) j
-    sys l a b s i j (i = i1) = l a .paths (b , s) j
-
-    square : ∀ (x : is-identity-system R r) a b (s : R a b)
-           → Square {A = Σ A (R a)}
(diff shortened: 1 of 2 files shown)


TOTBWF:
Nice! Mind if we add local Hedberg's theorem while we are at it? (IE: a constant map `f : x = y -> x = y` implies that `A` is a set). Should be super easy with `contr→identity-system`.

ncfavier:
Sure! Any idea why it's called local? Only reference I can find is https://www.cs.bham.ac.uk/~mhe/GeneralizedHedberg/html/GeneralizedHedberg.html.

ncfavier:
I think what you're asking for is `path-collapsible-is-hset` (condition (i) in the above link), and `local-hedberg` is just the "based" version of that. So I'm inclined to call it "generalised Hedberg's theorem", if anything.

TOTBWF:
The idea is that `f : ∀ {y : A} → x ≡ y → x ≡ y` is a "locally constant" map, as we've fixed the left basepoint.

ncfavier:
Right, what they call "local variant of Hedberg's theorem" in that paper is just the based version of Hedberg's theorem, and doesn't mention constant functions. Did you want me to prove that, or 2⇒1 in theorem 3.10?

TOTBWF:
> Actually I'm gonna go with "do it yourself" since I'm not sure what proof you had in mind 😄. Feel free to push to this branch

Way ahead of you, was just going to post a proof instead of an explanation 😁

TOTBWF:
After writing the proof out, it should probably live in another PR: it needs a page on coherently constant maps to really make sense. Will do that separately 🙂

ncfavier:
There's a bit about coherently constant maps in 1Lab.HIT.Truncation

TOTBWF:
Also, removed a now-redundant bit of `is-identity-system-is-prop`

TOTBWF:
> There's a bit about coherently constant maps in 1Lab.HIT.Truncation

Yeah, I found myself moving things around too much for what should be a little PR :)

TOTBWF (approved):

TOTBWF (dismissed):

PR #375, opened 2024-04-12
Simplex Category

TOTBWF:
# Description

This PR reworks the simplex category to make it more amenable to defining diagrams. It also proves a whole host of properties about it, which required the introduction of antiinjective and antisurjective functions. I was not able to find a better name for these, so suggestions are welcome!

Finally, this PR also introduces a bit of machinery for constructing concrete categories: a macro is used to introduce
top-level copattern matches so that we can get better goal display.

## Checklist

Before submitting a merge request, please check the items below:

- [x] I've read [the contributing guidelines](https://github.com/plt-amy/1lab/blob/main/CONTRIBUTING.md).
- [x] The imports of new modules have been sorted with `support/sort-imports.hs` (or `nix run --experimental-features nix-command -f . sort-imports`).
- [x] All new code blocks have "agda" as their language.

If your change affects many files without adding substantial content, and
you don't want your name to appear on those pages (for example, treewide
refactorings or reformattings), start the commit message and PR title with `chore:`.

PR #376, opened 2024-04-14
def: macros for creating copattern definitions

TOTBWF:
# Description

There are a couple of places in the 1Lab where we manually eta-expand out copattern definitions by hand to get better goal display. This mostly comes up when dealing with things like subcategories/forgetful functors; Agda will very happily unfold your nicely named category into `Restrict Blah`, which is not particularly helpful!

Manually performing these copattern expansions is a bit of a pain, so this PR adds a small macro that performs this mechanical busywork for us. This removes the need for the `declare-concrete-category` macro in #375.

## Checklist

- [x] The imports of new modules have been sorted with `support/sort-imports.hs` (or `nix run --experimental-features nix-command -f . sort-imports`).

### DIFF
--- a/src/1Lab/Reflection.lagda.md
+++ b/src/1Lab/Reflection.lagda.md
@@ -201,6 +201,10 @@ under-abs (lam v (abs nm _)) m = extend-context nm (arg (arginfo v (modality rel
 under-abs (pi a (abs nm _))  m = extend-context nm a m
 under-abs _ m = m
 
+extend-context* : ∀ {a} {A : Type a} → Telescope → TC A → TC A
+extend-context* [] a = a
+extend-context* ((nm , tm) ∷ xs) a = extend-context nm tm (extend-context* xs a)
+
 new-meta : Term → TC Term
 new-meta ty = do
   mv ← check-type unknown ty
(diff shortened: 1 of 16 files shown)


ncfavier:
Can this be unified with Cat.Functor.Coherence?

Lavenza:
[Pull request preview](https://preview.1lab.dev/376)

TOTBWF:
> Can this be unified with Cat.Functor.Coherence?

Should be able to! We will need to generalize `make-copattern` to work on functions into record types, but this shouldn't be too hard.

TOTBWF:
Should be good to go. Played around with a version of "deep" `cohere!` that would let us solve things like https://1lab.dev/Cat.Functor.Adjoint.html#25843 but this proved to be a bit tricky

plt-amy:
Looks good, are there places we can use it? (Sheaves, I guess? But I'm not done writing that yet :/)

TOTBWF:
Sorry, have been quite busy recently! Sheaves would be a good place, likewise for categories of free F-algebras

plt-amy (approved):

PR #377, opened 2024-04-30
chore: Fix typo

FernandoChu:
# Description

Simple typo.


ncfavier:
I don't think it's a typo: it's emphasizing funct*ion* as opposed to funct*or*.

FernandoChu:
Ohhhh my bad, though it's not actually italicizing it, not sure what the solution for that should be.

ncfavier:
Oh right, it seems like this only works with `funct*ion*`.

PR #396, opened 2024-06-19
web: don't justify the TOC

ncfavier:
There's not enough space for that to make sense.


Lavenza:
[Pull request preview](https://preview.1lab.dev/396)

PR #425, opened 2024-08-24
fixup type hover

plt-amy:
The current prettifying attempt generates some mangled types

so I think it's best to drop it and then try to fix this stuff upstream.


Lavenza:
[Pull request preview](https://preview.1lab.dev/425)

PR #431, opened 2024-09-11
Solver for equations over ℚ

plt-amy:
We can re-use the ring solver for expressions over ℚ, even involving inverses. The gist of it is that we can re-express any equation involving inverses to an equation between quotients of polynomials, then invoke the ring solver. Even though this idea seemed pretty dumb (surely there would be a cleverer algorithm over ℚ?), it's what the Rocq `field` tactic does as well. I also added some more comments explicating why ℚ is defined the way it is.


Lavenza:
[Pull request preview](https://preview.1lab.dev/431)

src/Prim/Data/Sigma.lagda.md
```diff
@@ -37,6 +37,10 @@ infixr 4 _,_
 
 syntax Σ-syntax X (λ x → F) = Σ[ x ∈ X ] F
 infix 4 Σ-syntax
+
+instance
+  make-Σ : ∀ {ℓ ℓ'} {A : Type ℓ} {B : A → Type ℓ'} ⦃ x : A ⦄ ⦃ y : B x ⦄ → Σ A B
```
TOTBWF, line 42: Where does this get used?

src/Data/Rational/Properties.lagda.md
```diff
@@ -0,0 +1,193 @@
...
+  /ℚ-cross {d = d} p = sym (ap (λ e → (e /ℚ d) ⦃ _ ⦄) (sym p) ∙ /ℚ-factorr)
+
+  /ℚ-same
+    : ∀ {q q' d d'} ⦃ α : Nonzero d ⦄ ⦃ β : Nonzero d' ⦄
+    → q *ℚ d' ≡ q' *ℚ d → q /ℚ d ≡ q' /ℚ d'
+  /ℚ-same p = /ℚ-cross (/ℚ-scaler ∙ sym (/ℚ-cross (sym p)))
+
+  /ℚ-frac
+    : ∀ {n d} ⦃ p : ℤ.Positive d ⦄
+    → (let instance _ = to-nonzero-frac (ℤ.positive→nonzero p))
```
TOTBWF, line 122: It feels like a bit of instance engineering should let us get rid of this?

src/Data/Rational/Solver.agda
```diff
@@ -0,0 +1,308 @@
...
+rational!-worker hole = withNormalisation false $ withReduceDefs (false , dr) $ do
+  goal ← infer-type hole >>= wait-for-type
+  just (lhs , rhs) ← get-boundary goal
+    where nothing → typeError $ strErr "Can't determine boundary: " ∷
+                                termErr goal ∷ []
+  elhs , vs ← normalise lhs >>= build empty-vars
+  erhs , vs ← normalise rhs >>= build vs
+  size , env ← environment vs
+
+  withReduceDefs (false , []) do
```
TOTBWF, line 279: This is a nice trick!

TOTBWF (dismissed):

src/Prim/Data/Sigma.lagda.md
plt-amy: Oh, for a bit I had instance search fill in the `Wf` constraints for `solve`, but now they're inferred from the expression (since I managed to make `invℚ` and `/ℚ` injective). Anyway, it feels useful to keep around if we end up in another situation where we want to solve for a tuple of constraints...

TOTBWF: Makes sense! And yeah, seems generally useful; only nit would be naming

src/Data/Rational/Properties.lagda.md
plt-amy: Well, it's impossible to do it completely without backtracking, but we can optimize for this case specifically.

src/Data/Rational/Solver.agda
plt-amy: :slightly_smiling_face:

src/Prim/Data/Sigma.lagda.md
plt-amy: I think our naming convention for instances means that this one should be called `Σ`, but, uh. Well

TOTBWF: There's a reason I didn't offer any alternatives 😁

TOTBWF (approved):

TOTBWF (dismissed):
numbers 👍

src/Data/Rational/Solver.agda
```diff
@@ -0,0 +1,309 @@
...
+
+open is-ring-hom
+
+{-
+Equational solver for expressions over the rational numbers. The
+strategy is as follows:
+
+- First, we reflect (accurately) a pair of rational expressions into a
+  value of 'Exp', defined below, which contains constructors for all the
+  basic operations on ℚ. Note that, since _*ℚ_ and _-ℚ_ are neutral on
```
Taneb, line 37: Do you mean to say `_*ℚ_` here? I don't think you compute multiplication to multiplication by the inverse

plt-amy: Yeah, that's what the comment is pointing out: `x *ℚ y` is neutral but `inc x *ℚ inc y` is some nonsense

PR #433, opened 2024-09-23
General adjoint functor theorem

plt-amy:
Proof from the nLab using joint equalisers and weakly initial objects. Only 3.3.2 and 3.3.3 from Borceux

### DIFF
--- a/src/Cat/Diagram/Equaliser/Joint.lagda.md
+++ b/src/Cat/Diagram/Equaliser/Joint.lagda.md
@@ -1,0 +1,172 @@
+<!--
+```agda
+open import Cat.Diagram.Limit.Base
+open import Cat.Prelude
+
+import Cat.Reasoning as Cat
+```
+-->
+
+```agda
+module Cat.Diagram.Equaliser.Joint {o ℓ} (C : Precategory o ℓ) where
+```
+
+<!--
+```agda
+open Functor
+open Cat C
+open _=>_
+```
+-->
+
+# Joint equalisers
+
+The **joint equaliser** of a family $(F_i : \hom(a, b))_i$ of parallel
+maps is a straightforward generalisation of the notion of [[equaliser]]
+to more than two maps. The universal property is straightforward to
+state: the joint equaliser of the $F_i$ is an arrow $e : E \to a$,
+satisfying $F_i e = F_j e$, which is universal with this property.
+
+```agda
+record is-joint-equaliser {ℓ'} {I : Type ℓ'} {E x y} (F : I → Hom x y) (equ : Hom E x) : Type (o ⊔ ℓ ⊔ ℓ') where
+  field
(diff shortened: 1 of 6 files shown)


src/Cat/Diagram/Equaliser/Joint.lagda.md
```diff
@@ -0,0 +1,172 @@
...
+    diagram .F₁ {false} {false} f = id
+    diagram .F-id {true} = refl
+    diagram .F-id {false} = refl
+    diagram .F-∘ {true} {true} {true} f g = introl refl
+    diagram .F-∘ {false} {true} {true} f g = introl refl
+    diagram .F-∘ {false} {false} {true} f g = intror refl
+    diagram .F-∘ {false} {false} {false} f g = introl refl
+```
+
+If the family is pointed, and the `diagram`{.Agda} has a limit, then
```
TOTBWF, line 144: Might be worth noting in a footnote that if the family is empty, then we get a product instead. Also could be good to point out that we could get away with merely inhabited, as `is-joint-equaliser` is a prop?

src/Cat/Diagram/Initial/Weak.lagda.md
```diff
@@ -0,0 +1,163 @@
...
+<!--
+```agda
+module _ {o ℓ} (C : Precategory o ℓ) where
+  open Cat C
+```
+-->
+
+A **weakly initial object** is like an [[initial object]], but dropping
+the requirement of uniqueness. Explicitly, an object $X$ is weakly
+initial in $\cC$, if, for every $Y : \cC$, there exists an arrow $X
```
TOTBWF, line 31:
```suggestion
initial in $\cC$, if, for every $Y : \cC$, there merely exists an arrow $X
```

src/Cat/Functor/Adjoint/AFT.lagda.md
```diff
@@ -0,0 +1,110 @@
...
+```
+-->
+
+```agda
+module Cat.Functor.Adjoint.AFT where
+```
+
+# The adjoint functor theorem
+
+The **adjoint functor theorem** states a sufficient condition for a
```
TOTBWF, line 21: It would probably be a good call to explain why we care about the solution set condition in the first place: this is a common point of confusion, as the naive AFT looks much simpler! I have some prose in a draft version of this result somewhere that we may be able to pilfer.

TOTBWF (changes requested):
That proof is always so slick. I do think that we need to add a bit of prose explaining _why_ the solution set matters, but other than that this looks great!

src/Cat/Functor/Adjoint/AFT.lagda.md
plt-amy: Isn't that the paragraph starting on line 62?

TOTBWF (approved):
🌰🥑🌊

Issue #407, opened 2024-07-12
Another irrelevant instance search bug


ncfavier:
While fixing https://github.com/plt-amy/1lab/issues/405, I found another seemingly unrelated bug with irrelevant instance metas: with the current Agda master, the module

```agda
open import 1Lab.Type
open import 1Lab.HLevel
open import 1Lab.HLevel.Closure

postulate
  foo : .(is-prop ⊥) → ⊤

bar : ⊤
bar = foo (hlevel 1)
```

results in the error

```
/home/n/git/1lab/src/wip/bug.agda:13,12-20
⊥ !=< (x : _T_14) → _S_17 x
when checking that the expression hlevel 1 has type is-prop ⊥
```

<details>
<summary>
Log with -vtactic:100 -vtc.instance:30
</summary>

```
Checking wip.bug (/home/n/git/1lab/src/wip/bug.agda).
The type of the FindInstance constraint isn't known, trying to find it again.
Found instance type head:  Number
no instance field candidates
instance candidates from signature for goal:
  (Number _A_7)
  {Number-Nat} length: 1
blocker:
  _A_7
mutual block:
  {bar}
Postponing possibly recursive instance search.
The type of the FindInstance constraint isn't known, trying to find it again.
Instance type is not yet known. 
Can't figure out target of instance goal. Postponing constraint.
The type of the FindInstance constraint isn't known, trying to find it again.
Found instance type head:  H-Level
no instance field candidates
instance candidates from signature for goal:
  H-Level _T_5 1
  {H-Level-pi, H-Level-⊤, H-Level-⊥, H-Level-pi', H-Level-sigma,
   H-Level-PathP, H-Level-Lift, H-Level-is-contr,
   H-Level-is-equiv} length: 9
blocker:
  any(_T_5, _r_8)
mutual block:
  {bar}
Postponing possibly recursive instance search.
findInstance 2: constraint: _10@13977009405847656758; candidates left: 9
findInstance 3: t = H-Level _T_5 1
{ checkCandidates _10@13977009405847656758
  target: H-Level _T_5 1
  candidates
  - H-Level-pi : {T.ℓ : Level} {T : Type T.ℓ} {ℓ : Level} {n : Nat}
                 {S : T → Type ℓ} ⦃ _ : {x : T} → H-Level (S x) n ⦄ →
                 H-Level ((x : T) → S x) n
  - H-Level-⊤ : {n : Nat} → H-Level ⊤ n
[...]

Issue #428, opened 2024-08-27
Yet another irrelevant instance bug


ncfavier:
Minimised from https://github.com/the1lab/1lab/pull/427#discussion_r1732698143

```agda
open import Agda.Builtin.Nat
open import Agda.Builtin.Equality

data _≤_ : Nat → Nat → Set where
  0≤x : ∀ {x} → 0 ≤ x
  s≤s : ∀ {x y} → x ≤ y → suc x ≤ suc y

instance
  s≤s' : ∀ {x y} → ⦃ x ≤ y ⦄ → suc x ≤ suc y
  s≤s' ⦃ x ⦄ = s≤s x

  x≤sucy : ∀ {x y} ⦃ p : x ≤ y ⦄ → x ≤ suc y
  x≤sucy {.0} {y} ⦃ 0≤x ⦄ = 0≤x
  x≤sucy {.(suc _)} {.(suc _)} ⦃ s≤s p ⦄ = s≤s (x≤sucy ⦃ p ⦄)

  {-# INCOHERENT x≤sucy #-}

Positive : Nat → Set
Positive n = 1 ≤ n

postulate
  z : (b : Nat) → .⦃ _ : Positive b ⦄ → Nat
  foo : ∀ n → .⦃ _ : Positive n ⦄ → z n ≡ 0

bar : ∀ n → ⦃ _ : Positive n ⦄ → z n ≡ 0
bar n@(suc _) ⦃ p ⦄ = foo n

{-
p != s≤s _54 of type 1 ≤ suc n₁
when checking that n is a valid argument to a function of type
(n₂ : Nat) .⦃ _ = z₁ : Positive n₂ ⦄ → z n₂ ≡ 0
-}
```

If the instance argument to `foo` is not marked irrelevant, then the error is

```
Failed to solve the following constraints:
  Resolve instance argument _58 : Positive (suc n₁)
  Candidates
    s≤s' : {x y : Nat} ⦃ _ : x ≤ y ⦄ → suc x ≤ suc y
    p : 1 ≤ suc n₁
    (stuck)
```

I *think* it's trying to compare `z n ⦃ p ⦄` with `z n ⦃ s≤s' ⦄`, which should succeed but fails (why?)

ncfavier:
We've banished irrelevance, so this is now irrelevant.

PR #475, opened 2025-03-07
chore: bump agda

plt-amy:
While working on elementary topoi, I got so mad that a module felt slower than it should, that I went to fix the compiler.

This bumps us past https://github.com/agda/agda/commit/9e0eae64b2a31750af4e30a0c52ca3cd63cc10ef so @ncfavier @TOTBWF @4e554c4c the next time you try to run our Agda you'll have to e.g.

```bash
nix-shell . --run 'agda --setup'
```

to write out the primitive modules even though we don't use them.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,6 +59,7 @@ jobs:
           NIX_BUILD_SHELL: bash
           build_command: |
             set -eu
+            agda --setup
             1lab-shake -j all
             eval "$installPhase"
 
--- a/default.nix
+++ b/default.nix
@@ -45,7 +45,7 @@ let
 
   deps = with pkgs; [
     # For driving the compilation:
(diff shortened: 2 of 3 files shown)


Lavenza:
[Pull request preview](https://preview.1lab.dev/475)

ncfavier:
This seems to fix https://github.com/the1lab/1lab/issues/407 (but not https://github.com/the1lab/1lab/issues/428).

plt-amy:
Kind of expected since https://github.com/the1lab/1lab/issues/428 is a consequence of the behaviour I describe in https://github.com/agda/agda/pull/7728#issuecomment-2695418542 which I'm working with the stdlib folks in https://github.com/agda/agda-stdlib/issues/2644 to excise.

Lavenza (approved):

PR #478, opened 2025-03-16
reflection.induction: handle erased constructors

ncfavier:
Inspired by https://github.com/cmcmA20/cubical-mini/pull/48. Also adds `work-on-types` so that the macro works at all when `--erasure` is used (see https://github.com/agda/agda/issues/6124).


Lavenza:
[Pull request preview](https://preview.1lab.dev/478)
##### Changed pages

  * [1Lab.Reflection](https://preview.1lab.dev/478/1Lab.Reflection.html) (+4, -0)

plt-amy (approved):

PR #498, opened 2025-05-17
Discreteness and H-Levels of W-types

TOTBWF:
# Description

This PR proves that a W-type `W A B` is discrete when `A` is discrete and `B x` is `Listable` for every `x`. I've also added
a characterisation of equality types of W-types. These should make arguments about decidable equality a bit easier to make for inductive types: instead of doing a giant $O(n^2)$ case bash, we can construct an equivalence with a W-type instead. Some automation on this front would be useful, but I've spent enough time shaving this yak already :)

## Checklist

- [x] The imports of new modules have been sorted with `support/sort-imports.hs` (or `nix run --experimental-features nix-command -f . sort-imports`).

### DIFF
--- a/src/Data/Wellfounded/W.lagda.md
+++ b/src/Data/Wellfounded/W.lagda.md
@@ -517,3 +517,170 @@ and the resulting contradiction.
                 apd (λ i → subtree (w=v i)) $
                 is-set→cast-pathp B (Discrete→is-set auto) (λ i → bx)
 ```
+
+## Path spaces of W-types
+
+We can also use W-types to give a generic characterisation of path
+spaces of inductive types.
+
+<!--
+```agda
+module WPath {ℓ ℓ'} {A : Type ℓ} {B : A → Type ℓ'} where
+```
+-->
+
+Typically, we prove results about path spaces of inductive types via
+**encode-decode** arguments. The general idea is that if a type `T` is
+an inductive type, then we can construct a type family `Code : T → T → Type`
+via recursion on `T` which describe the equality type between each
+pair of constructors. We then construct a pair of maps
+
+- `encode : ∀ (x y : T) → x ≡ y → Code x y`
+- `decode : ∀ (x y : T) → Code x y → x ≡ y`
+
+which translate between paths in `T` and our recursively defined family.
+The final step is to show `encode` and `decode` are inverses, which gives
+us an equivalence between paths in `T` and our alternative representation
+of the path space.
+
+Our characterisation of paths in W-types will follow a similar trajectory.
+We start by observing that a path `p : w ≡ v` between two trees `w, v : W A B`
+consists of the following data:
+
+- A path `label-path p : label w ≡ label v` between labels.
+- A path `f bw ≡ g bv` for every `bw : B (label w)` and `by : B (label v)`
+  that are connected by a `PathP`{.Agda} over `label-path p`.
+
+```agda
+  private
+    label-path : ∀ {w v : W A B} → w ≡ v → label w ≡ label v
+    label-path p = ap label p
+
+    subtree-path
+      : ∀ {w v : W A B}
+      → (p : w ≡ v)
+      → ∀ {bw : B (label w)} {bv : B (label v)}
+      → PathP (λ i → B (label-path p i)) bw bv
+      → subtree w bw ≡ subtree v bv
+    subtree-path p q = apd (λ i → subtree (p i)) q
+```
+
+We can then turn this observation on its head, and *define* our type of
+codes recursively as trees of paths between constructors whose branching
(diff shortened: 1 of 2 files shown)


Lavenza:
[Pull request preview](https://preview.1lab.dev/498)
##### Changed pages

  * [Data.Wellfounded.W](https://preview.1lab.dev/498/Data.Wellfounded.W.html) (+317, -4)
  * [1Lab.Type.Pi](https://preview.1lab.dev/498/1Lab.Type.Pi.html) (+7, -0)

src/Data/Wellfounded/W.lagda.md
```diff
@@ -378,3 +422,272 @@ and some stock facts about substitution.
...
+```
+
+If the two labels are distinct, then `w` and `v` must be distinct.
+
+```agda
+        (no x≠y) →
+          no (λ w=v → x≠y (Id≃path.from (ap label w=v)))
+```
+
+Conversely, suppose the two labels `x` and `y` are equal. Our next move
```
ncfavier, line 477: Not a fan of "conversely" here. Maybe "on the other hand" or "hence"?

src/Data/Wellfounded/W.lagda.md
```diff
@@ -378,3 +422,272 @@ and some stock facts about substitution.
...
+expects branches taken from `B x`, yet `g` expects branches taken from
+`B y`. We know that `x` and `y` are equal, but this isn't a judgmental
+equality, so some sort of transport is required. Luckily, we have
+anticipated this problem: by using inductive equality, we can simply
+pattern match on the proof that `x ≡ᵢ y`, so we only
+need to consider the case where `x` and `y` are judgmentally equal.
+
+```agda
+        (yes reflᵢ) →
+          case holds? (∀ bx → f bx ≡ g bx) of λ where
```
ncfavier, line 492: Is this using the instance we're defining recursively? That's pretty cool, but maybe worth pointing out explicitly?

src/Data/Wellfounded/W.lagda.md
```diff
@@ -378,3 +422,272 @@ and some stock facts about substitution.
...
+<!--
+```agda
+module WPath {ℓ ℓ'} {A : Type ℓ} {B : A → Type ℓ'} where
+```
+-->
+
+Typically, we prove results about path spaces of inductive types via
+**encode-decode** arguments. The general idea is that if a type `T` is
+an inductive type, then we can construct a type family `Code : T → T → Type`
+via recursion on `T` which describe the equality type between each
```
ncfavier, line 542:
```suggestion
via recursion on `T` which describes the equality type between each
```

src/Data/Wellfounded/W.lagda.md
```diff
@@ -378,3 +422,272 @@ and some stock facts about substitution.
...
+      → (p : w ≡ v)
+      → ∀ {bw : B (label w)} {bv : B (label v)}
+      → PathP (λ i → B (label-path p i)) bw bv
+      → subtree w bw ≡ subtree v bv
+    subtree-path p q = apd (λ i → subtree (p i)) q
+```
+
+We can then turn this observation on its head, and *define* our type of
+codes recursively as trees of paths between constructors whose branching
+factor is given by the type `PathP`s over the constructor paths.
```
ncfavier, line 577: Is this missing a word?

src/Data/Wellfounded/W.lagda.md
```diff
@@ -378,3 +422,272 @@ and some stock facts about substitution.
...
+    (x , f) ≡ (y , g)
+      ≃˘⟨ Iso→Equiv Σ-pathp-iso ⟩
+    Σ[ p ∈ (x ≡ y) ] PathP (λ i → B (p i) → W A B) f g
+      ≃˘⟨ Σ-ap-snd (λ p → funext-dep≃) ⟩
+    Σ[ p ∈ (x ≡ y) ] (∀ {bw bv} → PathP (λ i → B (p i)) bw bv → f bw ≡ g bv)
+      ≃⟨ Σ-ap-snd (λ p → Π-impl-cod≃ λ bw → Π-impl-cod≃ λ bv → Π-cod≃ (λ q → Path≃Code (f bw) (g bv))) ⟩
+    Σ[ p ∈ (x ≡ y) ] (∀ {bw bv} → PathP (λ i → B (p i)) bw bv → Code (f bw) (g bv))
+      ≃⟨⟩
+    Code (sup x f) (sup y g)
+      ≃∎
```
ncfavier, line 602: Cool!

TOTBWF:
I think there is value in having both, but mostly because I think the proof of `Path≃Code` is cute 🙂 

We should probably provide a version of your identity system as an extensionality transformer though.

plt-amy (approved):

PR #510, opened 2025-06-16
Add KaTeX worker pool

### LINKED ISSUE #115: Build daemon for katex
I've been spending some time looking at what the slow bits are of a clean 1lab build:

| Name       | Count | Total  | Average | Max   |
|------------|-------|--------|---------|-------|
| node       | 1,931 | 6m18s  | 0.20s   | 0.53s |
| pdflatex   | 258   | 4m16s  | 0.99s   | 1.80s |
| agda       | 1     | 4m04s  | 4m04s   | 4m04s |
| agda html  | 227   | 48.81s | 0.22s   | 1.12s |
| git        | 383   | 12.58s | 0.03s   | 0.13s |
| pdftocairo | 258   | 9.54s  | 0.04s   | 0.22s |

While there's some things which are always going to be slow (<sub><sup>Agda</sup></sub>), the obvious issue here is KaTeX. During a fresh build, we're spinning up 1.9k KaTeX processes, each of which runs for a fraction of a second!

It would be worth experimenting with creating some sort of build daemon here; just a basic JS program which reads an equation from stdin and writes the resulting SVG to stdout. Our Shake code can then maintain a pool of these, avoiding the cost of spawning a process for each job.

TOTBWF:
# Description

Closes #115. On my machine, this turn an 11 minute `1lab-shake all -j --skip-agda` clean build into a much snappier 3 minutes.

### DIFF
--- a/support/shake/app/Shake/KaTeX.hs
+++ b/support/shake/app/Shake/KaTeX.hs
@@ -162,26 +162,21 @@ katexRules = versioned 3 do
 
   _ <- versioned 3 $ addOracleCache \(LatexEquation (display, tex)) -> do
     pre <- askOracle (ParsedPreamble ())
-    result <-
-      traced "katex" $
-      withKatexWorker workerQueue \KatexWorker{..} ->
-      handle (\(SomeException e) -> pure (T.pack $ displayException e)) $ do
-        -- [HACK: File Separator Control Characters].
-        -- Instead of trying to do some fiddly JSON streaming, we opt to
-        -- use a little known feature of ASCII to delimit our requests.
-        -- The control character 0x1C in ASCII encodes a file-separator,
-        -- which is left up to applications to interpret. We can rely on this never showing up
-        -- in our JSON, so it is safe to use this to delimit requests.
-        let job = Aeson.encode (encodeKatexJob display (applyPreamble pre tex)) <> "\FS"
-        LBS.hPutStr katexWorkerIn job
-        hFlush katexWorkerIn
-        -- We can't use 'LBS.takeWhile (0x1c /=)' here, as the output can contain utf8-encoded
-        -- text. To avoid this, we first decode to utf8.
-        LT.toStrict <$> LT.takeWhile ('\FS' /=) <$> LT.decodeUtf8With T.strictDecode <$> LBS.hGetContents katexWorkerOut
-    case T.stripPrefix "Ok:" result of
-      Just html -> pure (T.strip html)
-      Nothing -> error (T.unpack result)
-
+    traced "katex" $ withKatexWorker workerQueue \KatexWorker{..} -> do
+      -- [HACK: File Separator Control Characters].
+      -- Instead of trying to do some fiddly JSON streaming, we opt to
+      -- use a little known feature of ASCII to delimit our requests.
+      -- The control character 0x1C in ASCII encodes a file-separator,
+      -- which is left up to applications to interpret. We can rely on this never showing up
+      -- in our JSON, so it is safe to use this to delimit requests.
+      let job = Aeson.encode (encodeKatexJob display (applyPreamble pre tex)) <> "\FS"
+      putStrLn ("Writing job: " <> show job)
+      LBS.hPutStr katexWorkerIn job
+      hFlush katexWorkerIn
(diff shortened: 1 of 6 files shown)


Lavenza:
[Pull request preview](https://preview.1lab.dev/510)

plt-amy:
using a channel of processes to handle the pooling is super clever

support/shake/app/Shake/Utils.hs
```diff
@@ -23,6 +28,33 @@ nodeCommand opts = command (opts ++ [AddPath [] ["node_modules/.bin"]])
...
+#endif
+      nodeLibPath =
+#ifdef NODE_LIB_PATH
+        NODE_LIB_PATH
+#else
+        "node_modules/"
+#endif
+  in
+  (Process.proc "node" (path:opts))
+  { Process.env = Just [("PATH", nodeBinPath <> ":" <> "$PATH"), ("NODE_PATH", nodeLibPath)]
```
ncfavier, line 55: `$PATH` is not expanded by anything, but it doesn't matter because `PATH` is not used by anything either (you're not invoking a shell).

```suggestion
  { Process.env = Just [("NODE_PATH", nodeLibPath)]
```

support/shake/app/Shake/KaTeX.hs
```diff
@@ -58,31 +79,112 @@ getPreambleFor = askOracle . LatexPreamble
...
-    Stdout out <- nodeCommand [StdinBS stdin] "katex" args
-    pure . Text.stripEnd . Text.decodeUtf8 $ out
-
+    traced "katex" $ withKatexWorker katexPool \KatexWorker{..} -> do
+      -- [HACK: File Separator Control Characters].
+      -- Instead of trying to do some fiddly JSON streaming, we opt to
+      -- use a little known feature of ASCII to delimit our requests.
+      -- The control character 0x1C in ASCII encodes a file-separator,
+      -- which is left up to applications to interpret. We can rely on this never showing up
+      -- in our JSON, so it is safe to use this to delimit requests.
```
ncfavier, line 177: Not that it matters but NUL (0x00) is a bit more standard for this purpose (see the `-0`, `-z` or `--zero` flag of every GNU command-line utility that outputs text).

support/shake/app/Shake/Utils.hs
ncfavier: Do we really need to say where `node` is here? Wouldn't it just be in `PATH` if we're in a Nix shell?

TOTBWF: I suppose not; I added this before I realized that I needed to explicitly add `nodejs` as a dep in the nix script.

support/shake/app/Shake/KaTeX.hs
TOTBWF: Seems reasonable to me, added in https://github.com/the1lab/1lab/pull/510/commits/1ff6e9fec99b93b01e525643ca97c647713deb58.

TOTBWF:
Ok, it looks like some text is a bit messed up: see https://preview.1lab.dev/510/Order.Base.html#partially-ordered-sets

support/shake/app/Shake/KaTeX.hs
```diff
@@ -58,31 +79,139 @@ getPreambleFor = askOracle . LatexPreamble
...
+  }
+
+-- | Spawn a single katex-worker process.
+-- If we cannot spawn the process for whatever reason, 'spawnLatexWorker' throws an exception via 'fail'.
+spawnKatexWorker :: IO KatexWorker
+spawnKatexWorker =
+  Process.createProcess workerSpec >>= \case
+  (Just katexWorkerIn, Just katexWorkerOut, _, katexProcessHandle) -> do
+    -- We are going to be delimiting requests with the file separator control
+    -- char (See [HACK: File Separator Control Characters]), so we can eek out a bit
```
SquidDev, line 100: I guess this needs updating too, now that it uses `\0`?

SquidDev (approved):
One tiny thing, but otherwise this looks good to me (even if Haskell's lazy IO hurts my brain)! Thank you so much for working on this — I think we all knew it would be an improvement, but I hadn't realised *this* dramatic.

plt-amy (approved):

PR #513, opened 2025-06-19
defn: conceptual proof that ΩS¹ ≃ Int

plt-amy:
I was experimenting with how difficult it would be to prove that `S¹` is a "universal symmetry" directly and the answer turns out to be not difficult at all so here's an attempt at replacing the current proof (that assumes some integers exist and constructs an isomorphism) with a new proof that just shows the universal property and derives the iso from that.

### DIFF
--- a/src/Homotopy/Space/Circle.lagda.md
+++ b/src/Homotopy/Space/Circle.lagda.md
@@ -20,6 +20,14 @@ open import Data.Int
 module Homotopy.Space.Circle where
 ```
 
+<!--
+```agda
+private variable
+  ℓ ℓ' : Level
+  X : Type ℓ
+```
+-->
+
 # Spaces: The circle {defines="circle"}
 
 The first example of nontrivial space one typically encounters when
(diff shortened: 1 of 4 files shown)


src/Homotopy/Space/Circle.lagda.md
```diff
@@ -113,172 +121,213 @@ always-loop = S¹-elim loop (double-connection loop loop)
...
-_arbitrary_ implementation of the integers: any type that satisfies
-their type-theoretic universal property. We call this a _HoTT take_: the
-integers are the homotopy-initial space with a point and an
-automorphism. What we'd like to do is prove that the loop space of the
-circle is _also_ an implementation of the integers, but it's non-trivial
-to show that it is a set _directly_.
+We will now calculate that the first [[loop space]] of the circle at the
+basepoint is *a* type of integers, i.e. it satisfies the [[universal
+property of the integers]]. First, we generalise the construction of
+`möbius`{.Agda} to turn equivalence on an arbitrary type into a type
```
ncfavier, line 127:
```suggestion
`möbius`{.Agda} to turn an equivalence on an arbitrary type into a type
```

src/Homotopy/Space/Circle.lagda.md
```diff
@@ -113,172 +121,213 @@ always-loop = S¹-elim loop (double-connection loop loop)
...
+family over `S¹`{.Agda}. Transport over this family will be the
+universal map $\Omega S^1 \to X$ associated with an equivalence $e : X
+\simeq X$ and basepoint $x_0 : X$.
```
ncfavier, line 130: I first read this as defining "transport over this family", not "the universal map".

```suggestion
family over `S¹`{.Agda}. Transport over this family will give the
universal map $\Omega S^1 \to X$ associated with an equivalence $e : X
\simeq X$ and basepoint $x_0 : X$.
```

src/Homotopy/Space/Circle.lagda.md
```diff
@@ -113,172 +121,213 @@ always-loop = S¹-elim loop (double-connection loop loop)
...
+  ΩS¹-elim P pr pl l = subst P (pathβ base l) attempt where
+    base≡ totl : S¹ → Type _
+    base≡ = equiv→family rotΩS¹
+    totl  = equiv→family (over→total rotΩS¹ pl)
 ```
 
-For the other case, we must
-provide a path $\rm{loop}^{-} \equiv \rm{loop}^{-}$ laying over the
-loop, which we can picture as the boundary of a square.  Namely,
+By rotating the basepoint (given by the method $pr : P\ \rm{base}$), we
```
ncfavier, line 241:
```suggestion
By rotating the basepoint (given by the method $\rm{pr} : P\ \rm{base}$), we
```

src/Homotopy/Space/Circle.lagda.md
```diff
@@ -113,172 +121,213 @@ always-loop = S¹-elim loop (double-connection loop loop)
...
+      (f (a ∙ loop) ≡ go (a ∙ loop))  ≃∎
+    where
+      go : _ → _
+      go l = equiv→action r l p
+```
 
-When we're at the base point, we already know what we want to do: we
-_have_ a function $\rm{loop}^{-} : \bb{Z} \to (\rm{base} \equiv
-\rm{base})$ already, so we can use that.
+### Induction for loops
```
ncfavier, line 209:
```suggestion
### Loop induction
```

src/Homotopy/Space/Circle.lagda.md
```diff
@@ -113,172 +121,213 @@ always-loop = S¹-elim loop (double-connection loop loop)
...
+a loop into `base≡`{.Agda} agree, where one is the identity and the
+other is the first projection from rotating our basepoint. This
+completes the proof of the elimination principle, and indeed of the
+universal property.
+
+```agda
+    pathβ : ∀ y l → path y (subst totl l (refl , pr)) ≡ to-base≡ y l
+    pathβ y = J (λ y l → path y (subst totl l (refl , pr)) ≡ to-base≡ y l)
+      (transport-refl refl)
+  ```
```
ncfavier, line 303:
`````suggestion
```
`````

File: src/Homotopy/Space/Circle.lagda.md [hunk 6]
```diff
@@ -113,172 +121,213 @@ always-loop = S¹-elim loop (double-connection loop loop)
...
+P(l)$. The family associated to this action will be called
+`totl`{.Agda}, and we will also need a name for the family associated to
+`rotΩS¹`{.Agda}.
 
 ```agda
-  decode : ∀ x → Cover x → base ≡ x
-  decode base = loopⁿ
+  ΩS¹-elim P pr pl l = subst P (pathβ base l) attempt where
+    base≡ totl : S¹ → Type _
+    base≡ = equiv→family rotΩS¹
```
ncfavier, line 237: If we just use `base ≡_` instead of `base≡`, `to-base≡` becomes trivial and `path` slightly simpler:

```agda
    path : ∀ y → totl y → base≡ y
    path = S¹-elim fst $ ua→ λ _ → ∙-filler _ loop
```

plt-amy: huh you're totally right. hang on

plt-amy: that never even occurred to me because this proof is from way deep in the trenches and when i was writing the prose around it i didn't even stop to consider the structure of what i had come up with that deeply. great simplification, thanks

src/Homotopy/Space/Circle.lagda.md
```diff
@@ -113,172 +121,200 @@ always-loop = S¹-elim loop (double-connection loop loop)
...
-  \arrow["{\rm{loop}\inv}"'{pos=0.1}, from=4-2, to=5-1]
-  \arrow["{\rm{refl}}"{pos=0.2}, from=4-4, to=5-5]
-  \arrow["{\rm{loop}^{\rm{unglue}(n)}}"{marking, allow upside down}, draw=none, from=0, to=1]
+  {\rm{base}} \&\& {\rm{base}}
+  \arrow["a", from=1-1, to=1-3]
+  \arrow["{\rm{refl}}"', from=1-1, to=3-1]
+  \arrow["{\rm{loop}}", from=1-3, to=3-3]
+  \arrow["{a\bullet\rm{loop}}"', from=3-1, to=3-3]
 \end{tikzcd}\]
-~~~
```
ncfavier, line 229: Did you mean to remove this line?

src/Homotopy/Space/Circle.lagda.md
```diff
@@ -113,172 +121,200 @@ always-loop = S¹-elim loop (double-connection loop loop)
...
+  ΩS¹-elim P pr pl l = subst P (pathβ base l) attempt where
+    totl : S¹ → Type _
+    totl = equiv→family (over→total rotΩS¹ pl)
 ```
 
-For the other case, we must
-provide a path $\rm{loop}^{-} \equiv \rm{loop}^{-}$ laying over the
-loop, which we can picture as the boundary of a square.  Namely,
+By rotating the basepoint (given by the method $\rm{pr} : P\
+\rm{base}$), we get a value in $P$, but its type appears to be way off.
```
ncfavier, line 240:
```suggestion
\rm{refl}$), we get a value in $P$, but its type appears to be way off.
```

ncfavier (approved):

PR #515, opened 2025-06-27
chore: fix inconsistencies in Limit.Base

jake-87:
The diagrams and prose seemed to more consistently use `{*}`, but changing all occurrences to `\top` would also be an option.

I also found it odd that `!` was named in the prose, but not labeled on the diagram.

### DIFF
--- a/src/Cat/Diagram/Limit/Base.lagda.md
+++ b/src/Cat/Diagram/Limit/Base.lagda.md
@@ -102,9 +102,9 @@ Luckily, we can! If we take a step back, we can notice that we are
 trying to construct a map into a functor. What are maps into functors?
 Natural transformations! Concretely, let $D : \cJ \to \cC$ be some
 diagram.  We can encode the same data as a cone in a natural
-transformation $\eps : {!x} \circ \mathord{!} \to D$, where $!x : \top
+transformation $\eps : {!x} \circ \mathord{!} \to D$, where $!x : \{*\}
 \to \cC$ denotes the constant functor that maps object to $x$ and every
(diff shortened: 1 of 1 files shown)


Lavenza:
[Pull request preview](https://preview.1lab.dev/515)
##### Changed pages

  * [Cat.Diagram.Limit.Base](https://preview.1lab.dev/515/Cat.Diagram.Limit.Base.html) (± 5)

jake-87:
Ugh, the `!` label is on the wrong side. It would be greatly appreciated if someone with more tikz knowledge would be able to fix that 😅

ncfavier:
I think using `\top` would be better. Could you also change `!x` to `!_x`?

You can switch the side of the label by adding a `'` after the label, as in `"!"'`.

src/Cat/Diagram/Limit/Base.lagda.md
```diff
@@ -102,7 +102,7 @@ Luckily, we can! If we take a step back, we can notice that we are
 trying to construct a map into a functor. What are maps into functors?
 Natural transformations! Concretely, let $D : \cJ \to \cC$ be some
 diagram.  We can encode the same data as a cone in a natural
-transformation $\eps : {!x} \circ \mathord{!} \to D$, where $!x : \top
+transformation $\eps : {!_x} \circ \mathord{!} \to D$, where $!_x : \top
 \to \cC$ denotes the constant functor that maps object to $x$ and every
```
ncfavier, line 106: While you're at it:

```suggestion
\to \cC$ denotes the constant functor that maps every object to $x$ and every
```

PR #520, opened 2025-07-17
chore: typos and other small fixes

ncfavier:
Mostly a (late) review of https://github.com/the1lab/1lab/pull/517, with some other small fixes.


.gitignore
```diff
@@ -2,7 +2,7 @@
 
 
 # Canonical location for work-in-progress stuff
-src/wip/
```
plt-amy, line 5: no. this is where my wips are

PR #523, opened 2025-07-21
chore: fix typo

4e554c4c:
pretty sure that "being orthogonal to an object is equivalent to being orthogonal to an object" is a typo. feel free to correct me if im wrong here. i also moved like 2 units of punctuation.


ncfavier (approved):

plt-amy:
well, it's certainly not wrong

PR #530, opened 2025-07-27
chore: bump Agda

plt-amy:
Includes the fixes to @ncfavier's highlighting issue and @4e554c4c's `record where` issue and profiling/benchmarking options so we can be sad about the performance of parametrised modules and not much else


Lavenza:
[Pull request preview](https://preview.1lab.dev/530)
##### Changed pages

  * [Data.Fin.Base](https://preview.1lab.dev/530/Data.Fin.Base.html) (+6, -1)

PR #540, opened 2025-08-04
misc fixes

ncfavier:


.envrc
```diff
@@ -0,0 +1,5 @@
+watch_file support
```
TOTBWF, line 1: -1 on commiting `.envrc` files. I use a different setup for `direnv`, and this would leave me with a permanent unstaged change which is really annoying. Maybe add this as `.envrc-defaults` and add a note in `CONTRIBUTING.md` that users can run `echo 'source_env_if_exists .envrc-defaults' | tee -a .envrc && direnv allow` to use the suggested direnv setup?

ncfavier: I'll just drop it. Maybe you could convince me to use your setup instead.

TOTBWF: Big difference is that I use
```
nix_direnv_manual_reload
```

instead of `watch_file`. Makes switching back to old branches not set off a nix timebomb lol

TOTBWF (approved):

plt-amy (approved):

PR #542, opened 2025-08-09
Locally graded categories

TOTBWF:
# Description

Just adds the definition of locally graded categories. Further theory requires a bunch of coherence lemmas about bicategories which I don't really want to do right now :)

### DIFF
--- a/src/Cat/LocallyGraded/Base.lagda.md
+++ b/src/Cat/LocallyGraded/Base.lagda.md
@@ -1,0 +1,131 @@
+---
+description: |
+  Locally graded categories.
+---
+<!--
+```agda
+open import Cat.Bi.Base
+open import Cat.Prelude
+```
+-->
+```agda
+module Cat.LocallyGraded.Base where
+```
+
+<!--
+```agda
+private variable
+  ob ℓb₁ ℓb₂ : Level
+```
+-->
+
+# Locally graded categories {defines=locally-graded-category}
+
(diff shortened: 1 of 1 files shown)


src/Cat/LocallyGraded/Base.lagda.md
```diff
@@ -0,0 +1,131 @@
...
+For a displayed category $\cE$, hom sets are replaced hom *families* indexed by
+some category $B$. The action of $B$ is a bit hard to see initially, but
+becomes painfully clear once you've gotten your hands dirty: it is given
+by transport of hom families.
+
+For an enriched category $\cC$, we replace hom sets with hom *objects* $\cC(X,Y) : \cV$ taken
+from some [[monoidal category]] $V$. The action of $V$ is a bit tricky
+to see at first, but becomes clear when we consider generalized elements
+$\cV(\Gamma, \cC(X,Y))$ of a hom object: it is given via precomposition with
+a morphism $\cC(\Delta, \Gamma)$.
```
ncfavier, line 40: So are you saying that the set of Γ-graded morphisms from X to Y is taken to be V(Γ, C(X, Y))? Seems like this should be stated explicitly. I guess you also meant "precomposition with a morphism V(Δ, Γ)"? Also, "bit tricky to see at first" sounds a bit repetitive with the previous paragraph.

ncfavier (approved):

PR #538, opened 2025-08-03
wip: solver for CCCs

plt-amy:
turns out you can just use the infrastructure for evaluating λ-terms to evaluate maps in a free CCC and everything works out

PR #544, opened 2025-08-14
STLC refactor

ncfavier:
Based on https://github.com/the1lab/1lab/pull/538

- Merge `Cat.CartesianClosed.Lambda` into `Cat.CartesianClosed.Solver`, which now only talks about formal morphisms
- Define STLC over a λ-signature instead in `Cat.CartesianClosed.Free.Lambda`

The solver module is a bit of a Frankenstein's creature now. I didn't want to delete too much of the prose, but feel free to cut it down if you feel that it's redundant now. There is also the option of leaving those two modules as they were before and have two versions of the STLC: one as an "internal language" and one over a theory.


src/index.lagda.md
```diff
@@ -351,14 +351,12 @@ open import Cat.Bi.Diagram.Monad.Spans
 
 ### Cartesian closed categories
 
-To a type theorist, a [[Cartesian closed]] category is exactly a
-simply-typed lambda calculus. To cement this connection, we have an
-implementation of normalisation-by-evaluation for STLC, whose
-correctness is established relative to its semantics in an arbitrary
-CCC:
+<!-- TODO -->
```
ncfavier, line 354: 😖

PR #545, opened 2025-08-16
Ordinals

finegeometer:
I got nerd-sniped into implementing ordinals. I hope this is useful!

This is my first time contributing to a project, so please tell me if I'm doing anything wrong.

There are a few things that bother me.
- I chose to include both the strict and nonstrict orders in the definition of ordinal, rather than deriving the latter from the former. This is useful, in that it allows the nonstrict order to *definitionally* equal a preexisting order. But I'm still not convinced it was the right choice. What do you think?
- I went into this with the intention to implement ordinals, and that's what I did. But some of the theory generalizes to extensional well-founded orders that need not be transitive. If we care to provide that generalization, that might force a rewrite.
- This implementation of ordinals isn't particularly conducive to providing an `is-well-order` predicate, since the underlying type is bundled deeply into the `Ordinal` structure.
- Naming things is always hard, and I don't know whether I did a good job.
- I chose to index ordinals by a single universe level, since that felt most natural. An alternative would have been to use two universe levels, analagous to `Poset`. I don't know which choice is better.

### DIFF
--- a/src/Data/Ordinal/Base.lagda.md
+++ b/src/Data/Ordinal/Base.lagda.md
@@ -1,0 +1,668 @@
+<!--
+```agda
+open import 1Lab.Reflection.Record
+open import 1Lab.Prelude
+
+open import Data.Wellfounded.Properties hiding (<-wf)
+open import Data.Wellfounded.Base
+
+open import Order.Instances.Nat
+open import Order.Base using (Poset)
+
+import Data.Nat.Order as Nat
+import Data.Nat.Base as Nat
+```
+-->
+
+```agda
+module Data.Ordinal.Base where
+```
+
+# Ordinals {defines="well-order ordinal"}
+
+An **ordinal** captures the concept of a [[well-founded]] order.
+Such an order is necessarily a *strict* order (like `_<_`, not `_≤_`),
+since a well-founded relation cannot be reflexive.
+
+The conditions for a well-order an be expressed fairly concisely.
+A relation `_<_ : X → X → Type` is a well order
+if all of the following hold:
+
+* It is propositional.
+
+* It is well-founded.
+
+* It is transitive.
+
+* Whenever `{z | z < x} ≡ {z | z < y}`, we have `x ≡ y`.
+
+Every ordinal is in fact a poset,
+where `x ≤ y` iff `∀ {z} → z < x → z < y`.
+We choose to use a somewhat redundant definition of ordinal,
+where this poset is part of the *structure*,
+since it often lines up with a preexisting order.
+
+```agda
+record Ordinal (ℓ : Level) : Type (lsuc ℓ) where
+  no-eta-equality
+  field
+    poset : Poset ℓ ℓ
+  open Poset poset public
+  field
+    _<_ : Ob → Ob → Type ℓ
+    <-thin : ∀ {x y} → is-prop (x < y)
+    <-wf : Wf _<_
+    <→≤ : ∀ {x y} → x < y → x ≤ y
+    ≤-transforms-< : ∀ {x y} → (x ≤ y) ≃ ∀ {z} → z < x → z < y
+  <-trans : ∀ {x y z} → x < y → y < z → x < z
+  <-trans x<y y<z = fst ≤-transforms-< (<→≤ y<z) x<y
+```
+
+<!--
+```agda
+  <-irrefl : ∀ {x} → ¬ (x < x)
+  <-irrefl {x} x<x = go (<-wf x) where
+    go : ¬ Acc _<_ x
+    go (acc h) = go (h x x<x)
+  <-asym : ∀ {x y} → x < y → y < x → ⊥
+  <-asym x<y y<x = <-irrefl (<-trans x<y y<x)
+
+private variable ℓ ℓ₁ ℓ₂ ℓ₃ : Level
+
+instance
+  Underlying-Ordinal : Underlying (Ordinal ℓ)
+  Underlying-Ordinal = record { ⌞_⌟ = Ordinal.Ob }
+```
+-->
+
+For example, the ordinal `ω` is given by the natural numbers,
+under their standard ordering.
+
+```agda
+ω : Ordinal lzero
+ω = record where
+  poset = Nat-poset
+  _<_ = Nat._<_
+  <-thin = Nat.≤-is-prop
+  <→≤ = Nat.<-weaken
+  ≤-transforms-< {x} {y} = prop-ext
+    Nat.≤-is-prop (hlevel 1)
+    (λ x≤y z≤x → Nat.≤-trans z≤x x≤y)
+    λ h →
+      let
+        h' : {z : Nat} → z Nat.≤ x → z Nat.≤ y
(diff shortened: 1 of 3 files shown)


Lavenza:
[Pull request preview](https://preview.1lab.dev/545)
##### New pages

  * [Data.Ordinal.Base](https://preview.1lab.dev/545/Data.Ordinal.Base.html) (+668, -0)
##### Changed pages

  * [HoTT](https://preview.1lab.dev/545/HoTT.html) (+36, -0)
  * [1Lab.Path.IdentitySystem](https://preview.1lab.dev/545/1Lab.Path.IdentitySystem.html) (+12, -0)

src/Data/Ordinal/Base.lagda.md
```diff
@@ -0,0 +1,668 @@
...
+  no-eta-equality
+  private
+    module α = Ordinal α
+    module β = Ordinal β
+  field
+    pres-< : ∀ {x y} → x α.< y → f x β.< f y
+  restriction : (x : α.Ob) → ⌞ Subordinal α x ⌟ → ⌞ Subordinal β (f x) ⌟
+  restriction x (y , y<x) = f y , pres-< y<x
+  field
+    restriction-is-equiv : ∀ {x} → is-equiv (restriction x)
```
TOTBWF, line 295: I think the following API might be easier to use, as we don't need `Equiv.to` and `Equiv.from` everywhere, and we don't have to unpack pairs.

```agda
record is-simulation
  (α : Ordinal ℓ₁) (β : Ordinal ℓ₂)
  (f : ⌞ α ⌟ → ⌞ β ⌟) : Type (ℓ₁ ⊔ ℓ₂) where
  no-eta-equality
  private
    module α = Ordinal α
    module β = Ordinal β
  field
    pres-< : ∀ {x y} → x α.< y → f x β.< f y
    sim : ∀ {x y} → y β.< f x → ⌞ α ⌟
    simulates : ∀ {x y} → (y<fx : y β.< f x) → f (sim y<fx) ≡ y
    sim-< : ∀ {x y} → (y<fx : y β.< f x) → sim y<fx α.< x

  reflect-≤ :  ∀ {x y} → f x β.≤ f y → x α.≤ y
  reflect-≤ {x} {y} fx≤fy = go (α.<-wf x) (α.<-wf y) fx≤fy where
    go : ∀ {x y} → Acc α._<_ x → Acc α._<_ y → f x β.≤ f y → x α.≤ y
    go {x} {y} (acc hx) (acc hy) fx≤fy =
      Equiv.from α.≤-transforms-< λ {z} z<x →
        let fz<fy : f z β.< f y
            fz<fy = β.<-≤-trans (pres-< z<x) fx≤fy

            fy* : ⌞ α ⌟
            fy* = sim fz<fy

            fy*<y : fy* α.< y
            fy*<y = sim-< fz<fy

            fy*≡z : sim fz<fy ≡ z
            fy*≡z =
              α.≤-antisym
                (go (hy fy* fy*<y) (hx z z<x) (β.≤-refl' (simulates fz<fy)))
                (go (hx z z<x) (hy fy* fy*<y) (β.≤-refl' (sym (simulates fz<fy))))
        in subst (α._< y) fy*≡z fy*<y

  has-injective : injective f
  has-injective p = α.≤-antisym
    (reflect-≤ (β.≤-refl' p))
    (reflect-≤ (β.≤-refl' (sym p)))

  segment : (x : α.Ob) → ⌞ Subordinal α x ⌟ → ⌞ Subordinal β (f x) ⌟
  segment x (y , y<x) = f y , pres-< y<x

  segment-is-equiv : ∀ {x} → is-equiv (segment x)
  segment-is-equiv {x = x} .is-eqv (y , y<fx) .centre =
    (sim y<fx , sim-< y<fx) , Σ-prop-path! (simulates y<fx)
  segment-is-equiv {x = x} .is-eqv (y , y<fx) .paths ((z , z<x) , fz=y) =
    Σ-prop-path! $ Σ-prop-path! $ has-injective (simulates y<fx ∙ sym (ap fst fz=y))
```

src/Data/Ordinal/Base.lagda.md
```diff
@@ -0,0 +1,668 @@
...
+```
+
+## The ordinal of ordinals {defines="ord transfinite-induction"}
+
+The collection of ordinals is itself well-ordered,
+with the strict inequality coming from the notion of subordinal.
+Induction over this ordering is known as **transfinite induction**.
+
+```agda
+_<ₒ_ : Ordinal ℓ₁ → Ordinal ℓ₂ → Type (ℓ₁ ⊔ ℓ₂)
```
TOTBWF, line 542: Seems like this could be nicer if we defined the strict order on ordinals using something like
```
record is-bounded-simulation
  (α : Ordinal ℓ₁) (β : Ordinal ℓ₂)
  (f : ⌞ α ⌟ → ⌞ β ⌟) : Type (ℓ₁ ⊔ ℓ₂)
  where
  no-eta-equality
  private
    module α = Ordinal α
    module β = Ordinal β
  field
    has-simulation : is-simulation α β f

  open is-simulation has-simulation public

  field
    bound : ⌞ β ⌟
    reflect : ∀ (x : ⌞ β ⌟) → x β.< bound → ⌞ α ⌟
    reflect-simulation : is-simulation (Subordinal β bound) α (uncurry reflect)
```

This gives us a bit more control over `<ₒ→≤ₒ`, as we are just dropping the bound.

src/Data/Ordinal/Base.lagda.md
```diff
@@ -0,0 +1,668 @@
...
+  no-eta-equality
+  field
+    poset : Poset ℓ ℓ
+  open Poset poset public
+  field
+    _<_ : Ob → Ob → Type ℓ
+    <-thin : ∀ {x y} → is-prop (x < y)
+    <-wf : Wf _<_
+    <→≤ : ∀ {x y} → x < y → x ≤ y
+    ≤-transforms-< : ∀ {x y} → (x ≤ y) ≃ ∀ {z} → z < x → z < y
```
TOTBWF, line 56: Do we want this to be an equivalence? Seems like it might be more nicely packaged as 2 fields.

TOTBWF (changes requested):
Thanks so much for the contribution! I'll split up the review a bit to not make it too overwhelming :)

First round is focused on some big-picture design questions, proof cleanup, some automation tips. I totally agree that packaging up both the strict and non-strict orders is a good idea, but I'm a bit unsure about bisimulations as a bespoke definition.

anshwad10:
@finegeometer Have you taken a look at @LorenzoMolena 's formalization of ordinals in the cubical library? I noticed they define the strict order on the product [a bit differently than you did](https://agda.github.io/cubical/Cubical.Data.Ordinal.Base.html#6309). Regarding `_≃ₒ_`, I don't really see why it is necessary; Earlier I was formalizing ordinals in Arend and I was able to prove that ≤ of ordinals is antisymmetric just fine ([here it is in a gist](https://gist.github.com/anshwad10/9994e15f4333817c4584cfd74afecc1a), may be hard to read if you don't know Arend though)

Issue #474, opened 2025-03-05
`is-total-order` is not a property


TOTBWF:
Currently, the definition of `is-total-order` is 

```agda
record is-total-order {o ℓ} (P : Poset o ℓ) : Type (o ⊔ ℓ) where
  open Poset P public

  field
    compare         : ∀ x y → (x ≤ y) ⊎ (y ≤ x)
```

This is not a property: in a discrete order, this is equivalent to a function `compare : P -> P -> Bool`. This can be fixed by truncating the result of `compare`; note that we can still compute meets and joins, as those are properties!

ncfavier:
Can we use `(x ≤ y) ⊎ (y < x)` instead? That is a proposition and is probably usable for sorting? I'm not sure if it's equivalent to what we have now, though.

Seiryn21:
Is `src/Order/Total.lagda.md` the only file to change ? Because if so there is only 8 lines to change and 1 to add to use this new definition

PR #546, opened 2025-08-18
Homotopy.Join as Logical Disjunction

finegeometer:
Suppose `X` is a type, and we want to produce an element of `X` under the assumption "`P` or `Q`". Intuitively, we should have to provide an element under the assumption `P`, provide an element under the assumption `Q`, and show that the elements agree when both assumptions hold.

This sounds exactly like the elimination principle for `_∗_` from `Homotopy.Join`! And indeed, `P ∗ Q` is a proposition that correctly represents the logical disjunction of `P` and `Q`. I provide the groundwork required to use it this way.

This could be useful, for example, as a potential resolution to issue #474. But beyond that, I'd argue that `P ∗ Q` is a good replacement for `∥ P ⊎ Q ∥` pretty much anywhere. It's equivalent, but it lets you eliminate into arbitrary types rather than just propositions.

---

In the process of implementing this, I also prove that a join with a contractible type is always contractible. This fact should be useful on its own.

### DIFF
--- a/src/Homotopy/Join.lagda.md
+++ b/src/Homotopy/Join.lagda.md
@@ -143,6 +143,31 @@ join-commutative .snd = is-iso→is-equiv record where
   linv = Pushout-elim (λ _ → refl) (λ _ → refl) λ c i j → glue c i
 ```
 
+Finally, if either of the joined types is contractible, so is the join.
+```agda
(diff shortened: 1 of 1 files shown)


TOTBWF:
Yeah. If we were to do anything I think a bespoke logical disjunction type would be best, along with a proof that it has the universal property of a join. Would make for a good `Inductive` instance.

PR #551, opened 2025-08-20
Add a case bash tactic

TOTBWF:
# Description

This PR adds an instance-based `case-bash!` tactic, which lets us reduce goals like `(x : Bool) -> (y : Nat) -> ...` to 
an `All` of each of the pointwise goals. When combined with other proof automation like `decide!`, this lets us automate proofs that are big case bashes followed by all `refl`.

### DIFF
--- a/src/1Lab/Type/Pi/Currying.lagda.md
+++ b/src/1Lab/Type/Pi/Currying.lagda.md
@@ -1,0 +1,142 @@
+---
+description: |
+  Currying pi types.
+---
+<!--
+```agda
+open import 1Lab.Type.Sigma
+open import 1Lab.Type.Pi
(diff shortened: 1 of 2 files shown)


ncfavier:
Do we need `Uncurry` for anything? Can we have a single bidirectional typeclass with instances like `Curried ((a : A) (b : B a) → C a b) ((p : Σ A B) → uncurry C p)` instead?

TOTBWF:
I thought about it, but didn't really like how the code was looking. It also makes it more difficult to add instances that do things like remove `⊤` arguments, as you'd end creating a lot of competing overlapping instances.

plt-amy (approved):

PR #561, opened 2025-08-28
chore: remove Authors, rewrite CONTRIBUTING.md

plt-amy:
rip to anyone who's clicked the "i've read the contributing guidelines" checkbox.

oh i also removed the dependency graph because its novelty has totally worn off and even the comedy of it seizing up when you go to like Cat.Displayed.Base has also worn off.


CONTRIBUTING.md
```diff
@@ -1,27 +1,157 @@
...
+We welcome all contributions, but please understand that the 1Lab has a
+very small team of part-time maintainers. It might take a while for your
+PR to be reviewed. Adhering to the guidelines below will help the review
+process focus on the things that actually matter instead of wasting
+everyones' cycles on style quibbles.
 
 ## General guidelines
 
-British spelling **must** be used throughout: Homotopy fib**re**,
-fib**red** category, colo**u**red operad, etc --- both in prose and in
```
ncfavier, line 15: i think the novelty of this has also worn off — can we pick a (symmetric) convention?

plt-amy: we don't put spaces before any *other* punctuation...

ncfavier: (

ncfavier: i think we should follow an established convention instead of making up yet another one. this stuff isn't logical anyway

plt-amy: removing all the spaces is certainly going to cause merge conflicts with all the other open prs

ncfavier: i'm mostly asking for this to be recorded in the CONTRIBUTING file—enforcing the convention is another matter

ncfavier: thanks

ncfavier (approved):

PR #562, opened 2025-08-28
chore: clarify indentation style further

plt-amy:
Just after merging #561 I realised that there's a bit of @TOTBWF's style in particular (sorry!!) that always peeves me and I always fix but that I _completely_ forgot to codify against in the styleguide.

I _think_ the rules here in this patch make sense; I reformatted `Epireflective` to match. Let me know if you can find a counterexample that follows the rules-as-written but still looks stupid.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -313,7 +313,8 @@ column 136!)
 [dual of the modular law]: https://github.com/the1lab/1lab/blob/main/src/Cat/Allegory/Reasoning.lagda.md#L110
 
 In laid-out syntax, **two spaces of indentation should be used**,
-instead of alignment, whenever possible. There are some situations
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -378,6 +378,75 @@ foo = let x = 1
...
+f x =
+  record where
+    ...
+  where
+    ...
+```
+
+If the RHS is an application, even if the clause LHS is too long for any
+of it to fit on the same line, the first line **should** include the
+head function, and as many arguments as possible respecting the rules
```
ncfavier, line 415: Shouldn't the line length limit take precedence?

TOTBWF (approved):

ncfavier (approved):

PR #563, opened 2025-08-31
The regular hyperdoctrine of subobjects of a regular category

ncfavier:
- Images are stable in a regular category
- The regular hyperdoctrine of subobjects of a regular category
- Beck-Chevalley over pullback squares for the codomain fibration, hence the canonical comprehension structure on the codomain fibration has (very strong) comprehension coproducts

-------------

The API for images is a bit suboptimal: we define images in `Cat.Diagram.Image` in terms of universal morphisms for full subcategory inclusions into slice categories, but we really want mono-images to be defined in terms of subobject categories instead. Some concrete issues:

- There is some duplication between `Cat.Diagram.Image` and `Cat.Regular.Image`: both define the "same" universal property, for example.
- The cocartesian lifts `f ^! α` given by `Subobject-opfibration` do not definitionally compute to `Im (f ∘ α .map)`, because they have to pass through the `Cat.Diagram.Image` API; see `^!-is-Im`.

I don't really know how to improve the situation. Maybe we could reverse the dependency of `Cat.Displayed.Instances.Subobjects` on `Cat.Diagram.Image` so that the latter can define `Image` in terms of universal maps for `Slice→Sub`, but then we'd have to move the cocartesian stuff somewhere else. Ideas welcome.

### DIFF
--- a/src/Cat/Morphism/Strong/Epi.lagda.md
+++ b/src/Cat/Morphism/Strong/Epi.lagda.md
@@ -4,6 +4,7 @@ open import Cat.Diagram.Coequaliser.RegularEpi
 open import Cat.Diagram.Pullback.Properties
 open import Cat.Instances.Shape.Terminal
 open import Cat.Functor.FullSubcategory
+open import Cat.Morphism.Factorisation
 open import Cat.Diagram.Limit.Finite
 open import Cat.Morphism.Orthogonal
 open import Cat.Diagram.Equaliser
@@ -264,12 +265,11 @@ factorisation of $f$ on our hands!
 
 ```agda
 strong-epi-mono→image
-  : ∀ {a b im} (f : Hom a b)
-  → (a→im : Hom a im) → is-strong-epi a→im
-  → (im→b : Hom im b) → is-monic im→b
-  → im→b ∘ a→im ≡ f
+  : ∀ {a b} (f : Hom a b)
+  → Factorisation C StrongEpis Monos f
   → Image C f
-strong-epi-mono→image f a→im (_ , str-epi) im→b mono fact = go where
+strong-epi-mono→image f fac = go where
+  open Factorisation fac renaming (left∈L to str-epi; right∈R to mono)
   open Initial
   open /-Obj
   open /-Hom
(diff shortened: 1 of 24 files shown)


plt-amy:
> The cocartesian lifts f ^! α given by Subobject-opfibration do not definitionally compute to Im (f ∘ α .map), because they have to pass through the Cat.Diagram.Image API; see ^!-is-Im.

Is it possible, and if so is it worth it, to define the opfibration manually, to avoid this?

src/Cat/Morphism/Strong/Epi.lagda.md
```diff
@@ -264,12 +265,11 @@ factorisation of $f$ on our hands!
 
 ```agda
 strong-epi-mono→image
-  : ∀ {a b im} (f : Hom a b)
-  → (a→im : Hom a im) → is-strong-epi a→im
-  → (im→b : Hom im b) → is-monic im→b
-  → im→b ∘ a→im ≡ f
+  : ∀ {a b} (f : Hom a b)
+  → Factorisation C StrongEpis Monos f
```
TOTBWF, line 269: Much nicer, thank you!

src/Cat/Regular/Image.lagda.md
```diff
@@ -72,38 +79,323 @@ Im-universal f m {e = e} p = r where
...
+  Sub-regular .has-univalence x = Sub-is-category cat
+  Sub-regular .cartesian = Subobject-fibration
+  Sub-regular .cocartesian = Sub-opf
+  Sub-regular .fibrewise-meet = Sub-products lex.pullbacks
+  Sub-regular .fibrewise-top x = Sub-terminal
+  Sub-regular .subst-& f m n = Sub-is-category cat .to-path ^*-∩ₘ
+  Sub-regular .subst-aye f = Sub-is-category cat .to-path ^*-⊤ₘ
+```
+
+It remains to check the Beck-Chevalley condition and Frobenius
```
TOTBWF, line 315: Link to Beck-Chevalley?

TOTBWF (dismissed):

TOTBWF:
This looks good to me! Agree that we should think a bit about the Image API: maybe we could write out the definition in parts like we do for most other universal properties.

PR #564, opened 2025-08-31
chore: clean up `Cat.Displayed.Cartesian.Identity`

ncfavier:
Half of that module is already in `Cat.Displayed.Cartesian` and the other half doesn't need to be so long.


src/Cat/Displayed/Cartesian.lagda.md
```diff
@@ -665,6 +675,9 @@ there is exactly _one_ choice to be made, that is, no choice at all.
 Thus, we do not dwell on the distinction.
 
 :::{.definition #fibred-category alias="cartesian-fibration fibration"}
+A **Cartesian fibration** is a displayed category having Cartesian lifts
+for every right corner.
```
plt-amy, line 679:
```suggestion
A **Cartesian fibration** is a displayed category having [[Cartesian lifts]]
for every right corner.
```

(± line wrapping)

PR #589, opened 2026-01-25
Build system gardening 🪴

ncfavier:
**Warning**: boring PR.

This makes `cabal build 1lab-shake` succeed (without warnings), so that one can build the 1Lab outside of the provided Nix shell by following the instructions in the (updated) README, should one choose to inflict this upon oneself. To prevent further bitrot in `1lab-shake.cabal`, I also replaced the bespoke `mkDerivation` for building the shakefile with an equivalent `callCabal2nix`; this means that dependencies and other build options are only tracked in one place.

These changes are made with *four* use cases in mind:

- `nix-build` for building the 1Lab;
- `nix-shell --run '1lab-shake all'` for working on the 1Lab interactively;
- `nix-shell -A shakefile --run 'cabal run 1lab-shake -- all'` for  working on the 1Lab *and the build system* interactively;
- `cabal run 1lab-shake -- all` outside of a Nix shell, after following the instructions in the README for setting up dependencies.

Also removes `support/check.sh` as discussed in https://github.com/the1lab/1lab/pull/587, since the same thing can be achieved with `agda --build-library`.

Also bumps GHC to 9.10 to get dependencies from cache.nixos.org.

Also contains other trivial changes. The only breaking changes should be:
- The `shake` Cabal target is renamed to `1lab-shake`.
- `nix-shell -A shakefile` no longer exposes `ghcide` (possibly other stuff as well). If needed, this can be added in `shell.nix`. HLS is still included.

Let me know if you find anything else.

@4e554c4c Does this work for your use cases?


TOTBWF (approved):
We can deal with synchronizing the revs in the cabal file in later work, as it looks pretty non-trivial.

plt-amy (approved):

PR #605, opened 2026-03-13
Simplify and generalise functor solver

aathn:
# Description

The functor solver in Cat.Functor.Solver was fragile in multiple ways:
- It did not normalise expressions before attempting to solve
- It did not allow expressions involving other functors than the one specified

This PR attempts to improve it by:
- Normalising expressions before solving, similarly to Cat.Solver
- Generalising it to allow expressions involving multiple nested functor applications
- Simplifying the normalisation logic by reusing machinery from Cat.Solver

The new version of `functor!` takes a category as argument, and tries to solve the goal by distributing all functor applications it can see, as well as removing (functors applied to) identities and associating to the right.

## Checklist

### DIFF
--- a/src/Cat/Functor/Solver.agda
+++ b/src/Cat/Functor/Solver.agda
@@ -1,214 +1,153 @@
 open import 1Lab.Reflection
-open import 1Lab.Prelude
 
-open import Cat.Base
+open import Cat.Prelude
 
-open import Data.List
-
-import Cat.Reasoning as Cat
+import Cat.Reasoning as Cr
+import Cat.Solver as Cs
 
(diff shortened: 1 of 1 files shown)


plt-amy (changes requested):
I've got very minor quibbles, but the code overall looks very sensible. I pinged @TOTBWF for a review since he wrote the solver initially.

TOTBWF (dismissed):
Looks pretty good: storing the functors inside of the expression is clever. Only minor nit is with the reflection code: we should be able to re-use the patterns from the category solver.

plt-amy (dismissed):

plt-amy (approved):

PR #609, opened 2026-03-22
fixup: comprehension coproducts introduction rule

ncfavier:
The rule as currently written is the introduction rule for *Π*-types.


TOTBWF (approved):
Thanks for fixing this!

PR #615, opened 2026-04-02
defn: `∫-assoc`

ncfavier:
Making `Element` a `Σ` and proving that categories of elements interact well with the "slice of presheaves is presheaves on category of elements" equivalence.


plt-amy (approved):
🙏

PR #620, opened 2026-04-19
decidable/discrete/strongly total orders are the same

### LINKED ISSUE #474: `is-total-order` is not a property

ncfavier:
The first commit fixes https://github.com/the1lab/1lab/issues/474 by using join instead of sum in `is-total-order`. (This turned out to be more convenient than truncating.)

The second commit shows that decidable total orders, total orders with decidable equality, and our previous definition of total order (now renamed to *strong* total order) are in fact all equivalent!

### DIFF
--- a/src/Data/Sum/Base.lagda.md
+++ b/src/Data/Sum/Base.lagda.md
@@ -56,6 +56,14 @@ infix 0 if⁺_then_else_
 if⁺_then_else_ : A ⊎ B → C → C → C
 if⁺ inl _ then y else n = y
 if⁺ inr _ then y else n = n
+
+is-right : A ⊎ B → Bool
+is-right (inl _) = false
+is-right (inr _) = true
+
+is-left : A ⊎ B → Bool
+is-left (inl _) = true
+is-left (inr _) = false
 ```
 -->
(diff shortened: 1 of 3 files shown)


Lavenza:
[Pull request preview](https://preview.1lab.dev/620)
##### Changed pages

  * [Order.Total](https://preview.1lab.dev/620/Order.Total.html) (+255, -73)
  * [Order.Base](https://preview.1lab.dev/620/Order.Base.html) (+35, -1)
  * [Data.Sum.Base](https://preview.1lab.dev/620/Data.Sum.Base.html) (+8, -0)

plt-amy (approved):

PR #624, opened 2026-04-24
make Acc's argument an index instead of a non-regular parameter

ncfavier:
This aligns more closely with the induction principle we want for `Acc` and avoids relying on the ability to define non-regular parametrised data types.


Lavenza:
[Pull request preview](https://preview.1lab.dev/624)
##### Changed pages

  * [Data.Wellfounded.Base](https://preview.1lab.dev/624/Data.Wellfounded.Base.html) (± 2)

plt-amy (approved):

